PR #2, opened 2018-08-03
REST client utility

zachmullen:
The main purpose of this topic is to decide on two runtime libraries that we'll use for REST interaction, my suggestions being:

* [axios](https://github.com/axios/axios) for dealing with XHRs
* [js-cookie](https://github.com/js-cookie/js-cookie) for dealing with cookies

As you can see from the yarn.lock changes, they don't introduce any new transitive dependencies, and they're both light, modern, popular, and well-documented.

In addition to selecting these tools, there is a minimal REST class that I envision we will use via Vue's `provide/inject` framework, e.g. 

```
import RestClient from 'girder/rest';

...
// ancestor
provide: {
   girderRest: new RestClient()
}

...
// descendant
inject: ['girderRest'],
created() {
  this.girderRest.request.get('collection').then(...)
}
```

### DIFF
--- a/src/rest.js
+++ b/src/rest.js
@@ -1,0 +1,45 @@
+import axios from 'axios';
+import cookies from 'js-cookie';
+
+export default class RestClient {
+  constructor(apiRoot = '/api/v1', { useCookie = true }) {
+    this._axios = axios.create();
+    this._axios.defaults.baseURL = apiRoot;
+
+    if (useCookie) {
+      this.token = cookies.get('girderToken');
+    }
+  }
+
+  /**
+   * Register a callback function that will be called on error responses.
+   * @param {Function} fn The function to be called. The callback will receive
+   *     a single argument, an axios error object. If you wish to ignore the
+   *     error, return a value. If you wish to propagate the error, return
+   *     a rejected Promise with the original argument.
+   */
+  onErrorResponse(fn) {
+    this._axios.interceptors.response.use(undefined, fn);
+  }
+
+  /**
+   * Use this to make Web API requests.
+   * @returns An Object with the same API as axios.
+   */
+  get request() {
+    return this._axios;
+  }
+
+  get token() {
+    return this._token;
+  }
+
+  set token(val) {
+    this._token = val;
+    if (val) {
+      this._axios.defaults.headers.common['Girder-Token'] = val;
+    } else {
+      delete this._axios.default.headers.common['Girder-Token'];
+    }
+  }
+}
--- a/src/rest.js
+++ b/src/rest.js
@@ -1,34 +1,23 @@
-import axios from 'axios';
+import axios_ from 'axios';
 import cookies from 'js-cookie';
 
+/**
+ * This is a subclass of axios that is meant to add Girder-specific helper functionality.
+ */
 export default class RestClient {
-  constructor(apiRoot = '/api/v1', { useCookie = true }) {
-    this._axios = axios.create();
-    this._axios.defaults.baseURL = apiRoot;
+  constructor({ apiRoot = '/api/v1', token = cookies.get('girderToken'), axios = axios_.create() } = {}) {
+    Object.assign(this, axios, {
+      apiRoot,
+      token,
+    });
 
-    if (useCookie) {
-      this.token = cookies.get('girderToken');
-    }
-  }
-
-  /**
-   * Use this to make Web API requests.
-   * @returns An Object with the same API as axios.
-   */
-  get request() {
-    return this._axios;
-  }
-
-  get token() {
-    return this._token;
-  }
-
-  set token(val) {
-    this._token = val;
-    if (val) {
-      this._axios.defaults.headers.common['Girder-Token'] = val;
-    } else {
-      delete this._axios.default.headers.common['Girder-Token'];
-    }
+    this.interceptors.request.use(config => ({
+      ...config,
+      baseURL: this.apiRoot,
+      headers: {
+        'Girder-Token': this.token,
+        ...config.headers,
+      },
+    }));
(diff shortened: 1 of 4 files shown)


jcfr:
Do you also envision the `girder.rest.RestClient` to be available as a `girder-restclient-cli` on npm ?

zachmullen:
@jcfr good question, I'm not sure exactly what level of granularity we'll want in our distributed packages. It'd be a good discussion topic for the future.

jcfr:
> a good discussion topic for the future.

Agreed. Just wanted to mention the idea.

jbeezley:
I don't really like the idea of wrapping the axios API like this.  I would prefer the girder client object be an axios instance.  For example, you can simplify what I did [here](https://github.com/Kitware/resonantgeo/blob/master/src/rest.js#L72-L9) to extend axios.

Somewhat unrelated, but useful to keep in mind is what I did with the properties on that object (`apiRoot`, `token`, `user`, etc.).  These can be used as reactive properties without having to listen to events explicitly.  In the "girder" example [here](https://github.com/Kitware/resonantgeo/blob/master/examples/girder/src/App.vue#L87), I have several components that are watching the logged in user attached to the "session" (e.g. [here](https://github.com/Kitware/resonantgeo/blob/master/src/components/girder/GirderJobList.vue#L57-L92)). When changing the `apiRoot`, the user updates and every subcomponent refreshes it's data automatically.  I'm not totally married to the idea yet, but it has proved convenient in a few cases.

zachmullen:
The reason I exposed the axios instance as a property rather than the top-level object was to avoid any possible collisions between our own symbol names and axios symbol names, but if you think that's not worth worrying about, I can change it.

This is something that could be achieved as a mixin, but I actually wanted to make it usable even outside Vue-land, so I figured using it via `inject/provide` would make that straightforward and more safe and explicit in terms of namespacing.

jbeezley:
Nothing in that file depends on Vue.  What I did to inject the client object into Vue is similar to how vue-axios works, e.g. `Vue.prototype.$girder = new Session()`.

jbeezley:
> The reason I exposed the axios instance as a property rather than the top-level object was to avoid any possible collisions between our own symbol names and axios symbol names, but if you think that's not worth worrying about, I can change it.

I don't think it's a big deal, we are in control of what we add to the axios instance, so we just have to ensure that we don't override anything.  We benefit far more by not having to wrap every method it provides.

zachmullen:
> I don't think it's a big deal, we are in control of what we add to the axios instance, so we just have to ensure that we don't override anything.

That's easy to do right now, but maybe the next major version of axios would add some symbols that happen to be named the same as some symbols we've already added to our own API.

danlamanna:
My 2 cents, this is **exactly** what we did with the python girder-client which has made it much more difficult to use. We're taking a useful general API and wrapping it "just in case it changes" and will create an API that is more difficult to use in all cases. The only true benefit of this client will be that it appends a value to every request, something which is already trivial by axios and should be expected of anyone writing javascript applications which need to interact with a REST API.

zachmullen:
@danlamanna how is exposing the axios API as a property making it more difficult to use? It's really just a difference in notation:

`thing.get('some_url', ...)`

vs.

`this.subproperty.get('some_url', ...)`

Is that really "much more difficult to use"? It still exposes the entirety of the axios API, but also allows us to add our own helpers and state on top of it, all while guaranteeing the namespaces will never collide regardless of what happens upstream.

zachmullen:
@danlamanna what I've proposed [here](https://github.com/girder/web_client_next/pull/2/files#diff-62e69722264e5a578707565ff5fcf41cR19) is *not* what we did in the python client that you were arguing against on that old topic, it's precisely what I believe you're arguing we *should* do, which is to expose the underlying API rather than attempting to re-wrap it. Or am I misunderstanding what you were saying we should do?

src/rest.js
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * Use this to make Web API requests.
+   * @returns An Object with the same API as axios.
+   */
+  get request() {
+    return this._axios;
+  }
+
+  get token() {
+    return this._token;
```
kotfic: wouldn't  it be better to just return `this._axios.defaults.headers.common['Girder-Token'] = val` to prevent storing the data in two places?

brianhelba:
I'm going to schedule a meeting to review the roadmap for this repo.

Personally, I think some of the controversy around this PR can be elided if we remember that:
* Our goal is to produce reusable components for project needs that exist right now. Everything in this repo should facilitate that to some extent.
* We are not making compatibility guarantees, so if something doesn't pan out, we can always change the code.

So, I'd like to see two outcomes from our meeting:
* A list of the actual components that we have as a short / medium term goal. As a starting point, perhaps:
  - File upload widget
  - Login widget
  - Hierarchy browser widget
* Some mutual understanding of the balance between giving @zachmullen enough freedom to accomplish our primary goals without derailing every PR and the need to provide skeptical review early in the process before potential missteps become too entrenched.

zachmullen:
@jbeezley @brianhelba @danlamanna PTAL, I've changed the API as we discussed to be an axios subclass. The core is now essentially identical to the resonantgeo REST helper, though this does not implement any of the helper methods yet.

(One deviation I would suggest from the resonantgeo module would be that we do not bring in the `qs` package at all; the preferred way of sending `x-www-form-urlencoded` data in the body with axios is to pass a `FormData` object, so if we need a helper we can simply do:

```
export function formEncode(obj) {
  const fd = new FormData();
  Object.entries(obj).forEach(([k, v]) => {
    fd.set(k, v);
  });
  return fd;
}
```
)

PR #3, opened 2018-08-04
CircleCI integration

zachmullen:


.circleci/config.yml
```diff
@@ -0,0 +1,20 @@
...
+  build:
+    docker:
+      - image: circleci/node:10
+    steps:
+      - checkout
+      - restore_cache:
+          name: Restore Yarn Package Cache
+          keys:
+            - yarn-packages-{{ checksum "yarn.lock" }}
+      - run: yarn
```
brianhelba: This should maybe be `yarn install`, to make it more explicitly readable.

brianhelba: Actually, maybe not, as I see: https://github.com/girder/web_client_next/pull/1 now. If that's the way you want to go consistently, I'm fine with that.

zachmullen: I might be taking crazy pills, but I seem to remember when I typed `yarn install` I got a warning that I should just type `yarn`. However, when I do it now, I don't see it, and a Google search doesn't give any indication of which one is the "right" way.

PR #5, opened 2018-08-15
Upload module

zachmullen:


zachmullen:
I think we can cut this branch off here. Rather than creating a separate build for examples, I used the existing "app" build target as the way to show examples of the component, such that running `yarn serve` is what you'll need to do to run the examples.

I think on a subsequent branch, we should change the build scripts a little bit, such that the main `build` command, or maybe an alternative one, builds a new `lib` target pointing at an `index.js` entry that captures all the components in the library in a top-level module.

This should probably be reviewed commit by commit.

src/utils/s3.js
```diff
@@ -0,0 +1,106 @@
+import axios from 'axios';
+import { stringify } from 'qs';
+
+export default class S3Upload {
```
jbeezley: Could the logic in this class be replaced with https://github.com/TTLabs/EvaporateJS?  The answer might be no; I don't have any experience with S3 uploads.

zachmullen: I could see us maybe using that, however I don't think it would be much less source code for us, since we'd have to do a bunch of nonstandard configuration to get that library to work with Girder. Since the existing code works, I'd rather keep what's here for this PR, and then investigate using that library as a future topic.

zachmullen:
Anything blocking the merging of this branch? I'd like to get the Login/Register component done next if we approve of the current pattern.

jbeezley:
We should follow the naming conventions in Vue's [style guide](https://vuejs.org/v2/style-guide).
> Component names should always be multi-word, except for root App components.

src/utils/upload.js
```diff
@@ -0,0 +1,75 @@
...
+    if (this.upload.behavior && uploadBehaviors[this.upload.behavior]) {
+      return new uploadBehaviors[this.upload.behavior](this).start();
+    }
+    return this._sendChunks();
+  }
+
+  async resume() {
+    this.progress({ indeterminate: true });
+
+    if (this.upload.behavior && uploadBehaviors[this.upload.behavior]) {
```
jbeezley: When resuming a failed download (due to a quota error on DKC).  I got an error here `TypeError: Cannot read property 'behavior' of null`.

zachmullen:
> We should follow the naming conventions in Vue's style guide.

> > Component names should always be multi-word, except for root App components.

That's a reasonable rule. Any style rule that can be caught by eslint should be. Why isn't our eslint plugin `vue/essential` not catching this already, since that rule is listed as essential?

jbeezley:
I think the reason is because the component name is encoded in the file name rather than any specific text in the file.

zachmullen:
Does this mean we should add a `name` to each component? My understanding is that would only affect how components were used if they were registered globally, since using them as modules will allow them to be named whatever users want.

What should be our naming conventions to make sure we adhere to this rule? Vuetify uses `v-thing` with filenames `VThing`, we could similarly do `g-thing` / `GThing.vue`, or we could do something different. I don't have a strong preference.

jbeezley:
> Does this mean we should add a name to each component? My understanding is that would only affect how components were used if they were registered globally, since using them as modules will allow them to be named whatever users want.

No, I don't think we should add a name attribute; I just mean eslint cannot do validation of file names only the text in files.  Ensuring the name follows the guidelines needs to be a manual thing anyway.

> What should be our naming conventions to make sure we adhere to this rule? Vuetify uses v-thing with filenames VThing, we could similarly do g-thing / GThing.vue, or we could do something different. I don't have a strong preference.

I was thinking `girder-thing`, but I also don't care that much.  The guidelines generally prefer [avoiding abbreviations](https://vuejs.org/v2/style-guide/#Full-word-component-names-strongly-recommended).

zachmullen:
Since our components are written right now as single-file components that can only be used as modules, what do you guys think of just punting on this? I actually think just `Upload` is the best name for the *module* here in the context of its source code, but downstreams should probably use it as something else; to demonstrate that I can import it into our App example as

    import GirderUpload from './components/Upload';

I realize this might be an unpopular opinion, I'm just trying to make use of the greatness of the module system and the namespacing it provides naturally. That is, the point of namespaces is that you *don't* have to prefix every symbol underneath them, because the namespace does that for you.

jeffbaumes:
I agree that we could decouple file names from exported symbols. In the example of Vuetify, they (aggressively in my opinion) chose the `v-` prefix, and named files/directories that way, but there is this `index.ts` in which the component names (e.g. for `Vue.use(Vuetify)`) are actually defined, where we could change the filesystem layout without changing the exported names (or vice versa):

https://github.com/vuetifyjs/vuetify/blob/dev/src/components/index.js

FYI there are a few places I could find quickly where people are using `g-`, again aggressively IMO. The major conflict may come from people making Google-based libraries.

* https://github.com/phanan/vue-google-signin-button
* https://github.com/toshalinfotech/vue-g-components

zachmullen:
In both of those links @jeffbaumes they are adding the `name` field to their components, which is how they will be referenced if the elements are registered with Vue globally rather than modularly. If we do add the name field, I would suggest `girder-thing` like @jbeezley and @danlamanna wanted, but until we add that field, I think we should leave the filename as is and just change how we use it in the example App component.

jbeezley:
That works for me.

jeffbaumes:
I see, makes sense @zachmullen.

PR #7, opened 2018-08-25
Upload module testing

zachmullen:
This adds some unit testing of the upload module, and attempts to establish some practices surrounding mocking. I've chosen to use the [axios-mock-adapter](https://github.com/ctimmerm/axios-mock-adapter) package rather than the more first-class [moxios](https://github.com/axios/moxios) because it seems nicer to use and is also more popular.

Naturally, adding testing did uncover an actual bug, on which I'll comment inline.

### DIFF
--- a/src/utils/s3.js
+++ b/src/utils/s3.js
@@ -21,7 +21,7 @@ export default class S3Upload {
       partEl.appendChild(etagEl);
       root.appendChild(partEl);
     });
-    return new window.XMLSerializer().serializeToString(root);
+    return root.outerHTML;
   }
 
   async _multiChunkUpload() {
--- a/src/utils/upload.js
+++ b/src/utils/upload.js
@@ -22,7 +22,7 @@ export default class Upload {
     chunkLen = UPLOAD_CHUNK_SIZE,
   } = {}) {
     Object.assign(this, {
-      $rest, file, params, parent, progress, chunkLen, upload: null, offset: 0,
+      $rest, file, params, parent, progress, chunkLen, upload: null, offset: 0, behavior: null,
     });
   }
 
@@ -63,7 +63,8 @@ export default class Upload {
     }))).data;
 
     if (this.upload.behavior && uploadBehaviors[this.upload.behavior]) {
-      return new uploadBehaviors[this.upload.behavior](this).start();
+      this.behavior = new uploadBehaviors[this.upload.behavior](this);
+      return this.behavior.start();
     }
     return this._sendChunks();
   }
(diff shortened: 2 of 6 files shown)


src/utils/s3.js
```diff
@@ -21,7 +21,7 @@ export default class S3Upload {
       partEl.appendChild(etagEl);
       root.appendChild(partEl);
     });
-    return new window.XMLSerializer().serializeToString(root);
+    return root.outerHTML;
```
zachmullen: This was changed because whatever test browser environment mocha is using does not support `window.XMLSerializer`.

src/utils/upload.js
```diff
@@ -14,10 +14,15 @@ export default class Upload {
    * @param opts.progress {Function} A progress callback for the upload. It will receive an Object
    * argument with either ``"indeterminate": true``, or numeric ``current`` and ``total`` fields.
    * @param opts.params {Object} Additional parameters to pass on the upload init request.
+   * @param opts.chunkLen {Number} Chunk size for sending the file (integer number of bytes).
    */
-  constructor($rest, file, parent, { progress = () => null, params = {} } = {}) {
+  constructor($rest, file, parent, {
+    progress = () => null,
+    params = {},
+    chunkLen = UPLOAD_CHUNK_SIZE,
```
zachmullen: This was added simply so that we could unit test this using very small blobs. Maybe there will be some actual utility for it as well.

src/utils/upload.js
```diff
@@ -70,8 +76,8 @@ export default class Upload {
 
     this.progress({ indeterminate: true });
 
-    if (this.upload.behavior && uploadBehaviors[this.upload.behavior]) {
-      return new uploadBehaviors[this.upload.behavior](this).resume();
```
zachmullen: This is the bug I referred to in the PR description. The way I saw it manifest was that any time an S3 multipart upload was made to fail at any chunk beyond the first, or during the finalize step, the resume did not pick up properly and tried to start at the beginning. [We now remember the behavior state](https://github.com/girder/girder_web_components/pull/7/files#diff-8af87c0ce07db8082e10e648f84bda61R66) on the upload itself so that the state isn't lost each time `resume()` is called.

zachmullen: This made me think: we could get away with an even simpler API on the Upload class of just `start()` rather than `start()` and `resume()`. It's less expressive and powerful, but simpler and less constrained in how it should be used. If you guys prefer that, I'm happy to change it.

zachmullen:
One thing I'm looking for commentary on is the structure of `describe` and `it`. Is the nesting here appropriate? I did it that way because I need one layer of mocking for all of the test cases, and a second layer for a subset of them.

jbeezley:
I would have a slight preference to the tests for the s3 behavior in a different file.  I think if the implementation is decoupled enough for the source files to be separated, the tests should be as well.

zachmullen:
@jbeezley done

zachmullen:
@brianhelba this is merged, but it'd be great if you could take a look at it to provide your feedback.

PR #8, opened 2018-08-25
Build this library as a UMD module

zachmullen:
The `build` command now builds a UMD library target rather than the example app (which will still be what you see via `yarn serve`, though we can certainly add another script for building the example).

In addition to being able to import this into node or a downstream webpack build, this means the library can now be imported into the browser directly, as in the following example, which was working when I included the App.vue component in the library exports.

```
<meta charset="utf-8">
<title>girder demo</title>
<script src="https://unpkg.com/vue"></script>
<script src="./girder.umd.js"></script>
<link href='https://fonts.googleapis.com/css?family=Roboto:300,400,500,700|Material+Icons' rel="stylesheet">
<link rel="stylesheet" href="./girder.css">
<div id="app"></div>
<script>
const girderRest = new girder.RestClient({ apiRoot: 'http://localhost:8080/api/v1' });
girderRest.fetchUser().then(() => {
  new Vue({
    render: h => h(girder.App),
    provide: { girderRest },
  }).$mount('#app');
});
</script>
```

Note that Vue is externalized by design under vue-cli-service v3.

### DIFF
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -4,7 +4,6 @@ import 'vuetify/dist/vuetify.min.css';
 import Upload from './Upload.vue';
 
 Vue.use(Vuetify);
-Vue.config.productionTip = false;
 
 export default {
   Upload,
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -1,0 +1,11 @@
+import Vue from 'vue';
+import Vuetify from 'vuetify';
+import 'vuetify/dist/vuetify.min.css';
+import Upload from './Upload.vue';
+
+Vue.use(Vuetify);
+Vue.config.productionTip = false;
+
+export default {
+  Upload,
+};
--- a/README.md
+++ b/README.md
@@ -1,26 +1,17 @@
-# girder-lib
+# Girder web component library
 
-## Project setup
-```
-yarn
-```
+## For developers
+### Project setup
+    yarn
 
-### Compiles and hot-reloads for development
-```
-yarn serve
-```
+#### Compile and hot-reload for development
+    yarn serve
 
-### Compiles and minifies for production
-```
-yarn build
-```
+#### Build the library for production
+    yarn build
 
-### Lints and fixes files
-```
-yarn lint
-```
+### Lint and fix source code
+    yarn lint
 
-### Run your unit tests
-```
-yarn test:unit
-```
+### Run unit tests
+    yarn test:unit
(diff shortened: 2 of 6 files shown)


src/components/index.js
```diff
@@ -0,0 +1,11 @@
+import Vue from 'vue';
+import Vuetify from 'vuetify';
+import 'vuetify/dist/vuetify.min.css';
+import Upload from './Upload.vue';
+
+Vue.use(Vuetify);
+Vue.config.productionTip = false;
```
jbeezley: I don't think this will actually work in all cases because you are importing `Upload` before registering vuetify.  At least, I had issues trying to do this in ResonantGeo, I don't remember the specifics.

The only way I found to register plugins automatically for consumers is to create a Vue plugin that does it on `install`, .e.g. <https://github.com/Kitware/resonantgeo/blob/master/src/index.js#L22>.

This could also be why `productionTip` isn't working for you.

zachmullen: Hm, it worked fine for me with the UMD build, I was able to display the Upload component via the demo.html file shown above. Is it some kind of undefined behavior? Maybe based on when the templates get compiled?

Regarding the productionTip issue -- setting it to false here does work to remove the production tip, I just don't fully understand the behavior of the production tip. Am I supposed to have to add this line even when I'm building for production?

jbeezley: I'm trying to remember the details of the problem I saw.  I'll update if I figure it out.

Going by how [vue uses productionTip](https://github.com/vuejs/vue/blob/52719ccab8fccffbdf497b96d3731dc86f04c1ce/src/platforms/web/runtime/index.js#L63), it shouldn't be necessary if you are building for production (`NODE_ENV='production'`).  It's only relevant if you want to suppress the messages in development.

zachmullen: The UMD build is done in production mode (it even says so in the log output), and setting NODE_ENV for it doesn't change anything. I now see the source of the issue; the build on https://unpkg.com/vue is in development mode, hence the message. If I set it to https://unpkg.com/vue@2.5.17/dist/vue.min.js instead, it should be gone. I'll remove the productionTip directive from our code.

jbeezley: Doing some excavation of ResonantGeo, I found the issue that I actually had was that the Vue object included in the component library was different than the one included in the App.  When doing `Vue.use(resonantgeo)`, it got a reference to the correct object.  This issue is mitigated by the fact that Vue is not bundled into redistributed package by the new CLI, so it's probably okay as is.

PR #9, opened 2018-08-25
Npm packaging

zachmullen:


zachmullen:
@brianhelba can you mark approval or rejection of this PR, in particular based on the name field (i.e. are you ok with this taking over [the girder package namespace](https://www.npmjs.com/package/girder) currently held by the Backbone client library)?

@girder/developers on the off chance anyone else cares enough to opine :)

jbeezley:
Yeah, I was hesitant about the name.  Since we are essentially using "girder" as the component library now and `@girder/app` as the application, I think it makes sense.

zachmullen:
I don't think it's uncommon for a non-org namespace to host the "main" package for an ecosystem, and the org to host related packages, Vue does this for example.

brianhelba:
Given the churn and future uncertainty of the Girder client ecosystem, I'd suggest leaving the top-level `girder` name empty for now and unambiguously adding this as `@girder/components` or `@girder/lib`. We don't want downstreams to pin to `girder` as a canonical package name then later decide that we want to split or distribute packages further or in a different way.

zachmullen:
I agree that it would be desirable to host this in a less prominent name until we are more confident, but with that said:

> I'd suggest leaving the top-level girder name empty for now

The thing is, it's not currently empty, it currently contains the source files of the existing Girder (Backbone-based) library. Since this is the replacement for that same concept, I think it makes sense to put it in the same name.

But we don't have to do that right now. For the moment, I can put it under `@girder/something` until we can make a more long-term decision.

zachmullen:
@jbeezley @brianhelba PTAL, if you approve I will publish.

PR #10, opened 2018-09-03
Notification bus utility

zachmullen:
This is adapted from the ResonantGeo version

### DIFF
--- a/src/utils/notifications.js
+++ b/src/utils/notifications.js
@@ -20,7 +20,7 @@ export default class NotificationBus extends Vue {
    */
   constructor($rest, {
     EventSource = window.EventSource,
-    pollingInterval = [500, 5000, 500],
+    pollingInterval = [500, 5000, 1000],
     withCredentials = true,
     useEventSource = false,
     since = new Date(),
@@ -88,7 +88,7 @@ export default class NotificationBus extends Vue {
     }
   }
 
-  _poll(interval = this.pollingInterval[1]) {
+  _poll(interval = 0) {
     const [min, max, step] = this.pollingInterval;
     let nextInterval;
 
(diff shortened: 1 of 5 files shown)


src/rest.js
```diff
@@ -1,20 +1,30 @@
 import axios_ from 'axios';
 import cookies from 'js-cookie';
 import { stringify } from 'qs';
+import NotificationBus from './utils/notifications';
```
zachmullen: One of the downsides to this architecture is that it makes the REST helper module depend on Vue, even if downstreams don't need the notification bus support. We may want to consider alternative dependency relationships between the two modules.

src/rest.js
```diff
@@ -30,6 +40,28 @@ export default class RestClient {
         ...config.headers,
       },
     }));
+
+    if (notifications) {
+      this.notifications = new NotificationBus(this, notificationBusOpts);
```
jbeezley: I never really liked the bidirectional dependency here.  I wonder if it would be better to have `login`/`logout` events on this class to decouple the two would be better, e.g.
```javascript
class NotificationBus {
  constructor($rest, ...) {
    ...
    $rest.$on('login', this.connect);
    $rest.$on('logout', this.disconnect);
  }
}
```
These events could be used as an alternative to watching the `user` attribute directly.  Not that we would want to encourage doing so, but there are cases where it is hard to avoid.

It would make this class directly depend on Vue (or some alternate pubsub), but I'm not convinced that's necessarily a big problem.

src/utils/notifications.js
```diff
@@ -0,0 +1,118 @@
...
+      this._eventSource = null;
+      this.$emit('stop', this);
+    }
+  }
+
+  _poll(interval = 0) {
+    const [min, max, step] = this.pollingInterval;
+    let nextInterval;
+
+    this._poller = setTimeout(async () => {
```
jbeezley: I think this is good enough for now.  If we need a more robust implementation later, I would look into a an external library like https://www.npmjs.com/package/promise-poller or https://github.com/MathieuTurcotte/node-backoff or similar.

src/rest.js
zachmullen: I had also thought of that architecture, and I think it's better than the currently written one. For this PR, I may just back out cbd9ce8 and we can decide on this in a future PR scoped to just that topic.

zachmullen:
I'm happy with the behavioral nature of the unit tests now; I think this is ready for final review @jbeezley 

I plan to follow this up with a PR that will change the RestClient to support pubsub events so that RestClient and NotificationBus can be linked easily without having either one depend on the other statically.

PR #11, opened 2018-09-12
Make RestClient an event bus

zachmullen:
Depends on #10


zachmullen:
Rebased this now that #10 was merged; @jbeezley PTAL

PR #15, opened 2018-09-14
Upload widget design ideas

jeffbaumes:
Some ideas for simplification of design for the upload widget. @jtomeck 

image

### DIFF
--- a/src/components/Upload.vue
+++ b/src/components/Upload.vue
@@ -1,17 +1,15 @@
 <template lang="pug">
 .upload-wrapper
   slot(name="toolbar")
-    v-toolbar(dense, dark, color="success")
-      v-toolbar-title Upload
+    v-toolbar(dense, dark, flat, color="primary")
+      v-icon cloud_upload
+      v-toolbar-title Upload to
+        v-avatar
+          v-icon {{ ResourceIcons[dest._modelType.toUpperCase()] }}
+        | {{ dest.name }}
       v-spacer
       v-btn(v-if="closeable", icon, flat, @click="$emit('close')")
         v-icon close
-  .px-0.py-2
-    slot(name="header")
-      .title.px-3
-        v-avatar.mr-2(color="blue-grey lighten-4")
-          v-icon {{ ResourceIcons[dest._modelType.toUpperCase()] }}
-        span {{ dest.name }}
 
   slot(name="dropzone")
     .dropzone-wrapper(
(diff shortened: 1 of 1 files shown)


src/components/Upload.vue
```diff
@@ -1,66 +1,63 @@
...
 .upload-wrapper
   slot(name="toolbar")
-    v-toolbar(dense, dark, color="success")
-      v-toolbar-title Upload
+    v-toolbar(dense, dark, flat, color="primary")
+      v-icon cloud_upload
+      v-toolbar-title Upload to
+        v-avatar
+          v-icon {{ ResourceIcons[dest._modelType.toUpperCase()] }}
+        | {{ dest.name }}
```
zachmullen: My original reason for splitting this into two rows was that it's more responsive-friendly. In mobile environments, horizontal space is very limited but vertical space is not (since it's expected that scrolling vertically will happen but scrolling horizontally should not).

src/components/Upload.vue
```diff
@@ -1,66 +1,63 @@
...
-          v-progress-circular(v-if="file.status === 'uploading'", color="primary", :rotate="-90",
-              :value="progressPercent(file.progress)", :indeterminate="file.progress.indeterminate")
-          v-icon(v-if="file.status === 'done'", color="success", large) check
-          v-icon(v-if="file.status === 'error'", color="error", large) warning
-        v-list-tile-content
-          v-list-tile-title {{ file.file.name }}
-          v-list-tile-sub-title
-            span(v-if="file.progress.current") {{ formatSize(file.progress.current ) }} /&nbsp;
-            span {{ formatSize(file.file.size) }}
+  v-card
```
zachmullen: I've used `v-card` in this way in some other apps now, however it makes me a little uncomfortable because I don't understand the *semantics* of `v-card`. Vuetify has this to say about it:

> The v-card component is a versatile component that can be used for anything from a panel to a static image. The card component has numerous helper components to make markup as easy as possible. Components that have no listed options use Vue's functional component option for faster rendering and serve as markup sugar to make building easier.

For showing a box that has one or more of a title, image, description text, and action buttons, it makes sense, but I'm not sure whether it's a good idea to use it as though it was essentially a `div` tag; is it used here for styling on purpose?

I'm hoping @jtomeck can impart some wisdom on the subject.

src/components/Upload.vue
```diff
@@ -1,66 +1,63 @@
...
-        span {{ dest.name }}
 
   slot(name="dropzone")
     .dropzone-wrapper(
         v-if="!files.length", :class="dropzoneClass", @dragenter="dropzoneClass = 'animate'",
         @dragleave="dropzoneClass = null", @drop="dropzoneClass = null")
       .dropzone-message
         v-icon(size="50px") attach_file
-        .title.mt-3 {{ dropzoneMessage }}
+        .title {{ dropzoneMessage }}
```
zachmullen: I think this was a good use of the margin class, I think it looks better to have a little space between the icon and line of text.

src/components/Upload.vue
```diff
@@ -1,66 +1,63 @@
...
+  v-card
+    div(v-show="files.length && !errorMessage")
+      div(v-show="!uploading")
+        v-subheader {{ statusMessage }}
+        v-btn(@click="files = []") Clear all
+        v-btn(color="primary", @click="start") Start upload
+
+    div(v-if="errorMessage")
+      v-alert(:value="true", type="error") {{ errorMessage }}
+        v-btn(v-if="!uploading", color="purple darken-1", dark, @click="start")
```
zachmullen: In the spirit of #16 we should change this to a semantic color.

src/components/Upload.vue
jeffbaumes: Good point. I was having some trouble looking for an example of a subheader in vuetify/material but I'll move it back to its own line.

src/components/Upload.vue
jeffbaumes: Agreed. Was a bit overzealous here. I'll put it back.

src/components/Upload.vue
```diff
@@ -1,66 +1,63 @@
...
+          v-icon replay
```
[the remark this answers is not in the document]
jeffbaumes: Right, will do.

src/components/Upload.vue
zachmullen: Based on that description, this seems like a reasonable use.

src/components/Upload.vue
zachmullen: @jtomeck your thoughts? I also like a single line better, is there something clever we can do to get it on a single line but also convey what action is happening without using too much horizontal space?

jeffbaumes:
I went even more minimal, and went more all-in on setting this up as a card.

image

Now everything is in a `v-card`, and I used `v-card-title` and `v-card-actions` for the header and action buttons. There was a decent [example](https://vuetifyjs.com/en/components/cards#example-custom-actions) in Vuetify's Card docs that has a title and subtitle that I based this on. The rationale is that for a component like this, a card is probably best since it may be embedded in lots of contexts (not just places where it takes up the entire page). In those other contexts, there would already be a toolbar on the window or dialog, so it does not make sense to also have a toolbar in this component.

I removed `display: flex` in the container CSS because with it, it would inconsistently render the `v-card-title` padding whether it was in "dropzone" mode or not, causing the page to jitter (still no idea why but removing `display: flex` fixed it.

I also made the progress size status appear at the same location as the normal file size status message, and compute that in the computed prop. Seemed simpler and a more consistent look.

Would be great for @jtomeck to have a look and provide more feedback.

zachmullen:
I like it. Can we make the destination name stand out a bit? Bold, different color, new line, something like that?

src/components/Upload.vue
```diff
@@ -1,17 +1,20 @@
...
-      v-spacer
-      v-btn(v-if="closeable", icon, flat, @click="$emit('close')")
-        v-icon close
-  .px-0.py-2
-    slot(name="header")
-      .title.px-3
-        v-avatar.mr-2(color="blue-grey lighten-4")
-          v-icon {{ ResourceIcons[dest._modelType.toUpperCase()] }}
-        span {{ dest.name }}
+v-card.upload-wrapper
```
zachmullen: I think we should do `v-card.upload-wrapper(flat)` so that it doesn't get a border-radius and box-shadow on the root element.

src/components/Upload.vue
```diff
@@ -1,17 +1,20 @@
 <template lang="pug">
-.upload-wrapper
-  slot(name="toolbar")
-    v-toolbar(dense, dark, color="success")
```
zachmullen: If we are removing the toolbar, we may want to remove the `closeable` attribute and corresponding button since it may now be out of scope for this component.

src/components/Upload.vue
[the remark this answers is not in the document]
jtomeck: It appears that the alert content doesn't look vertically aligned properly (more space at the top than bottom. Vuetify's alert component usually has a darker 'border' element at the top. Is that missing from here (it could be why it doesn't look middle aligned to me)?

screen shot 2018-09-19 at 1 41 05 pm

jeffbaumes:
Pushed a small change to make the destination bold.

image

src/components/Upload.vue
jeffbaumes: +1 done.

src/components/Upload.vue
jeffbaumes: +1 done.

zachmullen:
With the `flat` property on our v-card, this looks nice inside a Dialog:

```
v-dialog(v-if="folder", max-width="800px")
    v-btn(slot="activator") open
    v-card
      girder-upload(v-if="folder", :dest="folder", :closeable="closeable", :multiple="multiple")
      v-card-actions
        v-btn(color="primary", flat) Close
```

screen shot 2018-09-19 at 1 50 43 pm

src/components/Upload.vue
```diff
@@ -72,10 +58,6 @@ export default {
   mixins: [sizeFormatter],
   inject: ['girderRest'],
   props: {
-    closeable: {
```
zachmullen: We'll need to modify `App.vue` to no longer pass this prop.

src/components/Upload.vue
```diff
@@ -1,17 +1,19 @@
...
-  .px-0.py-2
-    slot(name="header")
-      .title.px-3
-        v-avatar.mr-2(color="blue-grey lighten-4")
-          v-icon {{ ResourceIcons[dest._modelType.toUpperCase()] }}
-        span {{ dest.name }}
+v-card.upload-wrapper(flat)
+  v-card-title(primary-title)
+    div
+      slot(name="toolbar")
```
zachmullen: I think this slot should go *outside* the `v-card-title` to make it more powerful. I can imagine downstreams wanting to remove the top part altogether to add their own title, toolbar, whatever, e.g. to make an upload dialog where only the list part scrolls while the title remains fixed.

zachmullen: And we should rename the slot to "header" or something like that that corresponds to its semantics a bit more closely.

jeffbaumes: Done, and renamed slot to "header".

src/components/Upload.vue
jeffbaumes: Done.

src/components/Upload.vue
```diff
@@ -1,8 +1,8 @@
 <template lang="pug">
 v-card.upload-wrapper(flat)
-  v-card-title(primary-title)
-    div
-      slot(name="toolbar")
+  slot(name="header")
+    v-card-title(primary-title)
+      div
```
zachmullen: Is this `div` necessary?

jeffbaumes: Apparently so. `v-card-title` is set up as a horizontal flexbox (why? not sure) so direct child elements will be laid out horizontally. All Vuetify [examples](https://vuetifyjs.com/en/components/cards) that use `v-card-title` put all content within a single `div`.

jeffbaumes: Even a core developer states that it is a flexbox for "reasons unknown" https://github.com/vuetifyjs/vuetify/issues/3558

zachmullen:
Thanks for the improvements, I think it looks a lot nicer now and is generally more useful.

jtomeck:
@jeffbaumes - I think the screenshots look great, although I'd like to get it loaded up to look at it in full.  Once I do that I'll likely be able to give actual recommendations from there.

Issue #14, opened 2018-09-12
Have some process for auto-generating documentation


mgrauer:
xref https://github.com/girder/girder/pull/2563.

jeffbaumes:
This looks interesting

https://github.com/vue-styleguidist/vue-styleguidist

Since this is a component library, it would seem that the focus should be on great component docs.

There is also the Girder REST library and utils, which could perhaps generate a second doc site for this library with more standard JSDoc/ESDoc (with a main page pointing to "Components" and "API" docs), or it could be spun off into its own library, since this is the "components" repo.

jbeezley:
I spent some time looking into this for resonantgeo.  Many different options exist, but styleguidist seemed the most promising as a vue native tool.  I didn't use it at the time because it didn't support pug, but it looks like it has been [added](https://github.com/vue-styleguidist/vue-docgen-api/pull/49) recently.

zachmullen:
I just tried to use vue-styleguidist via its vue-cli plugin, but ran into [an issue](https://github.com/vue-styleguidist/vue-cli-plugin-styleguidist/issues/17) that I couldn't resolve, even in a barebones minimal reproduction. Hopefully we'll hear back on that, until then we can't document anything that uses `async/await`.

subdavis:
Worth looking into again, maybe.

zachmullen:
Good call. I don't know whether I made an earnest attempt on GWC after that upstream fix.

PR #17, opened 2018-09-14
Authentication UI

subdavis:
First take at a plug-n-play auth component.  I more or less stole @zachmullen's layout from algorithms.kitware.com and added OAuth.

screenshot_20180914_164809

I still want to...
* add autodocs with https://github.com/propellant/doctor
* Float the login screen on top of Upload widget if user is logged out.
* Continue to improve the README with project requirements, like mentioning the icon stylesheet setup and everything else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,72 +1,101 @@
 # Girder web component library
 
 ## For users
-### Installation
-    npm install @girder/components
-    
-or
 
-    yarn add @girder/components
+### Installation
+
+```bash
+npm install @girder/components
+# or
+yarn add @girder/components
+```
 
 ### Use as a prebuilt UMD module
 
 It's not necessary to install Girder web components yourself, you can import the prebuilt bundle
 into your page by including the following tags:
 
-    <script src="https://unpkg.com/vue/dist/vue.min.js"></script>
-    <script src="https://unpkg.com/@girder/components"></script>
-    <link rel="stylesheet" href="https://unpkg.com/@girder/components/dist/girder.css">
+```html
+<script src="https://unpkg.com/vue/dist/vue.min.js"></script>
+<script src="https://unpkg.com/@girder/components"></script>
+<link rel="stylesheet" href="https://unpkg.com/@girder/components/dist/girder.css">
+```
 
 This will expose all the library's components under the global variable `girder`, e.g.
 `girder.components.Upload`.
 
 ### Import components into your project
 
-If you're building your own downstream application, you can include individual components from
-the library. Either import the full UMD module:
+#### Girder RestClient
 
-    import girder from '@girder/components';
-    
-    const restClient = new girder.RestClient();
-    
+Many components in this library will require a `girder.RestClient` named `girderRest` through provide/inject.  
+The client can be provided through any common ancestor.  For example:
+
+```javascript
+import girder from '@girder/components';
+// apiRoot should point to your girder instance
+const restClient = new girder.RestClient({ apiRoot: 'http://localhost:8080/api/v1' });
+
+// Provide "girderRest" to the root
+new Vue({
+  provide: { girderRest },
+  ...
+}).$mount('#app');
+```
+
+#### Components
+
+If you're building your own downstream application, you can include individual components from the library. Because these are vuetify components, your consumer application is responsible for creating the `v-app` container.  [Read more](https://vuetifyjs.com/en/layout/pre-defined#all-about-app).  See `src/main.js` for a more comprehensive example.
+
+Either import the full UMD module:
+
+```javascript
+import girder from '@girder/components';
+```
+
 Or import specific components from the `src` directory:
 
-    // Downstream.vue
-    <template>
-      <girder-upload ... />
-    </template>
-    
-    <script>
-    import { Upload as GirderUpload } from '@girder/components/src/components';
-    
-    export default {
-      components: { GirderUpload },
-      ...
-    };
-    </script>
+```html
+<!-- Downstream.vue -->
+<template>
+    <girder-upload ... />
+</template>
+
+<script>
+import { Upload as GirderUpload } from '@girder/components/src/components';
+
+export default {
+    components: { GirderUpload },
+    ...
+};
+</script>
+```
 
 > **Note:** When importing components from the source tree, you should import
-  them from `index.js` rather than importing the `.vue` files yourself, as the
-  latter is prone to break if files get moved in future releases. For example:
-  ```
-  import { Upload as GirderUpload } from '@girder/components/src/components';  // Good
-  import GirderUpload from '@girder/components/src/components/Upload.vue'; // Unsafe -- may move in future
-  ```
-  > Files and symbols that do not appear in an index.js should be considered private and it
-  is unsafe to use them in downstream projects since they are not part of the supported API surface.
+> them from `index.js` rather than importing the `.vue` files yourself, as the
+> latter is prone to break if files get moved in future releases. For example:
+> ```javascript
+>  import { Upload as GirderUpload } from '@girder/components/src/components';  // Good
+>  import GirderUpload from '@girder/components/src/components/Upload.vue'; // Unsafe -- may move in future
+> ```
+> Files and symbols that do not appear in an index.js should be considered private and it
+> is unsafe to use them in downstream projects since they are not part of the supported API surface.
 
 ## For developers
-### Project setup
-    yarn
 
-#### Compile and hot-reload for development
-    yarn serve
+```bash
+# Project Setup
+yarn
 
-#### Build the library for production
-    yarn build
+# Compile and hot-reload for development
+yarn serve
 
-### Lint and fix source code
-    yarn lint
+# Build the library for production
+yarn build
 
-### Run unit tests
-    yarn test:unit
+# Lint and fix source code
+yarn lint
+
+# Run unit tests
+yarn test:unit
+```
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "axios-mock-adapter": "^1.15.0",
     "babel-core": "7.0.0-bridge.0",
     "babel-jest": "^23.6.0",
+    "pug-lint-vue": "^0.1.3",
     "vue-template-compiler": "^2.5.16"
   },
   "eslintConfig": {
--- a/package.json
+++ b/package.json
@@ -14,11 +14,7 @@
     "@mdi/font": "^2.8.94",
     "axios": "^0.18.0",
     "js-cookie": "^2.2.0",
-    "pug": "^2.0.3",
-    "pug-plain-loader": "^1.0.0",
     "qs": "^6.5.2",
-    "stylus": "^0.54.5",
-    "stylus-loader": "^3.0.2",
     "vue": "^2.5.16",
     "vuetify": "^1.1.9"
   },
@@ -33,6 +29,9 @@
     "babel-core": "7.0.0-bridge.0",
     "babel-jest": "^23.6.0",
     "pug-lint-vue": "^0.1.3",
+    "stylus": "^0.54.5",
+    "stylus-loader": "^3.0.2",
+    "vue-cli-plugin-pug": "^1.0.7",
     "vue-template-compiler": "^2.5.16"
   },
   "eslintConfig": {
--- a/package.json
+++ b/package.json
@@ -14,7 +14,11 @@
     "@mdi/font": "^2.8.94",
     "axios": "^0.18.0",
     "js-cookie": "^2.2.0",
+    "pug": "^2.0.3",
+    "pug-plain-loader": "^1.0.0",
     "qs": "^6.5.2",
+    "stylus": "^0.54.5",
+    "stylus-loader": "^3.0.2",
     "vue": "^2.5.16",
     "vuetify": "^1.1.9"
   },
@@ -28,10 +32,6 @@
     "axios-mock-adapter": "^1.15.0",
     "babel-core": "7.0.0-bridge.0",
     "babel-jest": "^23.6.0",
-    "pug-lint-vue": "^0.1.3",
-    "stylus": "^0.54.5",
-    "stylus-loader": "^3.0.2",
-    "vue-cli-plugin-pug": "^1.0.7",
     "vue-template-compiler": "^2.5.16"
   },
   "eslintConfig": {
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,5 +1,5 @@
 <template lang="pug">
-v-app.app
+v-app.app(v-if="ready")
   girder-upload(v-if="folder", :dest="folder", :multiple="multiple")
   v-layout(v-else, justify-center)
     v-flex(xs12, sm10, md8, lg6)
@@ -24,10 +24,11 @@ export default {
   },
   data() {
     return {
+      ready: false,
       error: null,
       folder: null,
       multiple: true,
-      forgotPasswordUrl: `${this.girderRest.url.origin}/#?dialog=resetpassword`,
+      forgotPasswordUrl: '/#?dialog=resetpassword',
     };
   },
   methods: {
@@ -51,12 +52,12 @@ export default {
       }
     },
   },
-  mounted() {
+  async mounted() {
     if (this.girderRest.user) {
-      this.fetchFolder();
+      await this.fetchFolder();
     }
+    this.ready = true;
     this.girderRest.$on('login', () => {
-      this.girderRest.fetchUser();
       this.fetchFolder();
     });
   },
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,12 +1,12 @@
 <template lang="pug">
-v-app#app
-  girder-upload(v-if="folder" :dest="folder", :multiple="multiple")
-  v-layout(v-else justify-center)
-    v-flex(xs12 sm10 md8 lg6)
+v-app.app
+  girder-upload(v-if="folder", :dest="folder", :multiple="multiple")
+  v-layout(v-else, justify-center)
+    v-flex(xs12, sm10, md8, lg6)
       girder-auth(
-        :register="true"
-        :oauth="true"
-        :forgot-password-url="forgotPasswordUrl")
+          :register="true",
+          :oauth="true",
+          :forgot-password-url="forgotPasswordUrl")
 </template>
 
 <script>
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,45 +1,41 @@
 <template lang="pug">
-v-app.app
-  girder-upload(v-if="folder", :dest="folder", :multiple="multiple")
-  v-form.pa-4(v-if="!girderRest.user")
-    v-text-field(v-model="username", label="Username or email")
-    v-text-field(v-model="password", label="Password", type="password")
-    v-alert(:value="error") {{ error }}
-    v-btn(@click="login") Login
+v-app#app
+  girder-upload(v-if="folder" :dest="folder", :multiple="multiple")
+  v-layout(v-else justify-center)
+    v-flex(xs12 sm10 md8 lg6)
+      girder-auth(
+        :register="true"
+        :oauth="true"
+        :forgot-password-url="forgotPasswordUrl")
 </template>
 
 <script>
-import { Upload as GirderUpload } from './components';
+import {
+  Upload as GirderUpload,
+  Authentication as GirderAuth,
+} from './components';
 
 export default {
   name: 'app',
   inject: ['girderRest'],
-  components: { GirderUpload },
-  data: () => ({
-    error: null,
-    folder: null,
-    multiple: true,
-    password: '',
-    username: '',
-  }),
+  components: {
+    GirderUpload,
+    GirderAuth,
+  },
+  data() {
+    return {
+      error: null,
+      folder: null,
+      multiple: true,
+      forgotPasswordUrl: `${this.girderRest.url.origin}/#?dialog=resetpassword`,
+    };
+  },
   methods: {
     _errMsgFromResp(response) {
       if (response) {
         this.error = response.data.message;
       } else {
         this.error = `Could not connect to server: ${this.girderRest.apiRoot}`;
-      }
-    },
-    async login() {
-      this.error = null;
-
-      try {
-        await this.girderRest.login(this.username, this.password);
-        this.username = '';
-        this.password = '';
-        await this.fetchFolder();
-      } catch ({ response }) {
-        this._errMsgFromResp(response);
       }
     },
     async fetchFolder() {
@@ -59,6 +55,10 @@ export default {
     if (this.girderRest.user) {
       this.fetchFolder();
     }
+    this.girderRest.$on('login', () => {
+      this.girderRest.fetchUser();
+      this.fetchFolder();
+    });
   },
 };
 </script>
--- a/src/components/Authentication/Authentication.vue
+++ b/src/components/Authentication/Authentication.vue
@@ -1,16 +1,15 @@
 <template lang="pug">
-.girder-authentication-component
-  v-card
-    v-tabs(v-model="activeTab", color="primary", slider-color="yellow", dark)
-      v-tab(key="login") Log In
-      v-tab(key="registration", v-if="register") Register
-      v-tab-item(key="login-box")
-        girder-login(:oauth-providers="oauthProviders",
-            :forgot-password-url="forgotPasswordUrl",
-            :forgot-password-route="forgotPasswordRoute",
-            @forgotpassword="$emit('forgotpassword')")
-      v-tab-item(key="registration-box", v-if="register")
-        girder-registration(:oauth-providers="oauthProviders")
+v-card.girder-authentication-component
+  v-tabs(v-model="activeTab", color="primary", slider-color="yellow", dark)
+    v-tab(key="login") Log In
+    v-tab(key="registration", v-if="register") Register
+    v-tab-item(key="login-box")
+      girder-login(:oauth-providers="oauthProviders",
+          :forgot-password-url="forgotPasswordUrl",
+          :forgot-password-route="forgotPasswordRoute",
+          @forgotpassword="$emit('forgotpassword')")
+    v-tab-item(key="registration-box", v-if="register")
+      girder-registration(:oauth-providers="oauthProviders")
 </template>
 
 <script>
--- a/src/components/Authentication/Authentication.vue
+++ b/src/components/Authentication/Authentication.vue
@@ -82,4 +82,5 @@ export default {
     padding 10px 16px
   .v-alert__icon
     margin-right 8px
+    margin-left 8px
 </style>
--- a/src/components/Authentication/Authentication.vue
+++ b/src/components/Authentication/Authentication.vue
@@ -36,7 +36,7 @@ export default {
     },
     /* A vue-router route path. */
     forgotPasswordRoute: {
-      type: [ Object, String ],
+      type: [Object, String],
     },
   },
   components: {
--- a/src/components/Authentication/Authentication.vue
+++ b/src/components/Authentication/Authentication.vue
@@ -67,20 +67,3 @@ export default {
   },
 };
 </script>
-
-<style lang="stylus">
-.girder-authentication-component
-  .v-btn
-    margin-left 0px
-    margin-right 16px
-  .v-icon.aligned.v-icon--left
-    margin-right 8px;
-  .v-form .container
-    padding-left: 24px;
-  .v-alert.error
-    margin 0px
-    padding 10px 16px
-  .v-alert__icon
-    margin-right 8px
-    margin-left 8px
-</style>
--- a/src/components/Authentication/Authentication.vue
+++ b/src/components/Authentication/Authentication.vue
@@ -5,19 +5,17 @@
       v-tab(key="login") Log In
       v-tab(key="registration", v-if="register") Register
       v-tab-item(key="login-box")
-        login(ref="loginForm",
-            :oauth-providers="oauthProviders",
+        girder-login(:oauth-providers="oauthProviders",
             :forgot-password-url="forgotPasswordUrl",
             :forgot-password-route="forgotPasswordRoute",
             @forgotpassword="$emit('forgotpassword')")
       v-tab-item(key="registration-box", v-if="register")
-        registration(ref="registerForm",
-            :oauth-providers="oauthProviders")
+        girder-registration(:oauth-providers="oauthProviders")
 </template>
 
 <script>
-import Login from './Login.vue';
-import Registration from './Register.vue';
+import GirderLogin from './Login.vue';
+import GirderRegistration from './Register.vue';
 
 export default {
   inject: ['girderRest'],
@@ -44,8 +42,8 @@ export default {
     },
   },
   components: {
-    Login,
-    Registration,
+    GirderLogin,
+    GirderRegistration,
   },
   data() {
     return {
--- a/src/components/Authentication/Authentication.vue
+++ b/src/components/Authentication/Authentication.vue
@@ -30,15 +30,13 @@ export default {
       type: Boolean,
       default: false,
     },
-    /* A full URL to be used as an anchor href to an external page.
-     * Overrides forgotPasswordRoute.
-     */
+    /* A full URL to be used as an anchor href to an external page. */
     forgotPasswordUrl: {
       type: String,
     },
-    /* A vue-router route path.  Must use object form. */
+    /* A vue-router route path. */
     forgotPasswordRoute: {
-      type: Object,
+      type: [ Object, String ],
     },
   },
   components: {
--- a/src/components/Authentication/Authentication.vue
+++ b/src/components/Authentication/Authentication.vue
@@ -1,5 +1,5 @@
 <template lang="pug">
-v-flex.pa-2(xs-12)
+v-flex.pa-2(xs-12)#girder-authentication-component
   v-card
     v-tabs(v-model='activeTab', color='primary', slider-color='yellow', dark)
       v-tab(key='login') Log In
@@ -74,14 +74,22 @@ export default {
 </script>
 
 <style lang="stylus">
+#girder-authentication-component
+  .no-decorate
+    text-decoration none
+  .v-btn
+    margin-left 0px
+    margin-right 16px
+  .v-icon.aligned
+    font-size 18px
+    vertical-align middle
+    padding-bottom 2px
+</style>
+
+<style lang="stylus" scoped>
 .v-alert.error
   margin 0px
   padding 10px 16px
 .v-alert__icon
   margin-right 8px
-.no-decorate
-  text-decoration none
-.v-btn
-  margin-left 0px
-  margin-right 16px
 </style>
--- a/src/components/Authentication/Authentication.vue
+++ b/src/components/Authentication/Authentication.vue
@@ -1,0 +1,87 @@
+<template lang="pug">
+v-flex.pa-2(xs-12)
+  v-card
+    v-tabs(v-model='activeTab', color='primary', slider-color='yellow', dark)
+      v-tab(key='login') Log In
+      v-tab(key='registration' v-if='register') Register
+      v-tab-item(key='login-box')
+        login(
+          ref='loginForm'
+          :oauth-providers='oauthProviders'
+          :forgot-password-url='forgotPasswordUrl'
+          :forgot-password-route='forgotPasswordRoute'
+          @forgotpassword='$emit("forgotpassword")')
+      v-tab-item(key='registration-box' v-if='register')
+        registration(
+          ref='registerForm'
+          :oauth-providers='oauthProviders')
+</template>
+
+<script>
+import Login from './Login.vue';
+import Registration from './Register.vue';
+
+export default {
+  inject: ['girderRest'],
+  props: {
+    register: {
+      type: Boolean,
+      default: false,
+      note: 'Enable registration from this component?',
+    },
+    oauth: {
+      type: Boolean,
+      default: false,
+      note: 'Enable OAuth login from this component?',
+    },
+    forgotPasswordUrl: {
+      type: String,
+      note: 'A full URL to be used as an anchor href to an external page.  Overrides forgotPasswordRoute.',
+    },
+    forgotPasswordRoute: {
+      type: Object,
+      note: 'A vue-router route path.  Must use object form.',
+    },
+  },
+  components: {
+    Login,
+    Registration,
+  },
+  data() {
+    return {
+      activeTab: null,
+      loginInProgress: false,
+      registerInProgress: false,
+      loginErrors: null,
+      registerErrors: null,
+      oauthProviders: [],
+    };
+  },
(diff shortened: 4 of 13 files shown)


brianhelba:
I think each public component should be importable directly from `src/component`. If that component has private sub-components, than the 'public component' should actually be a directory with an `index.js` file and a default export of the public component. Private sub-components associated with that should be in the same directory.

This has the benefit of clustering all files related to a given component in the same directory, which should be more maintainable than throwing many private sub-components together in a common `subcomponents` folder.

README.md
```diff
@@ -15,3 +16,26 @@
...
+Vue.use(Veutify);
+Vue.use(VeeValidate);
+
+new Vue({
+  render: h => h(App),
+  provide: { girderRest },
+}).$mount('#app');
+```
+
+Because these are veutify components, your consumer application is responsible for creating the `v-app` container.  [Read more](https://vuetifyjs.com/en/layout/pre-defined#all-about-app).
```
jeffbaumes: veutify => Vuetify

README.md
```diff
@@ -15,3 +16,26 @@
 
 ### Run unit tests
     yarn test:unit
+
+## Using this library
+
+To use components in your Vue.js app, the Vue instance must meet some requirements.  The basic requirements are:
+
+```javascript
+Vue.use(Veutify);
```
jeffbaumes: Veutify => Vuetify

src/utils/mixins.js
```diff
@@ -18,3 +18,22 @@ export const sizeFormatter = {
...
+    async validate() {
+      return this.$validator.validateAll();
+    },
+  },
+}
+
+export {
+  sizeFormatter, 
+  validator,
+}
```
jeffbaumes: Please change editor settings so that they add final newline to files automatically on save to avoid these.

README.md
```diff
@@ -15,3 +16,60 @@
...
+## Using this library
+
+Package requirements, to be added after `vue create ...`
+
+```bash
+# Vuetify
+vue add vuetify
+
+# Vee Validate
+yarn add vee-validate
```
zachmullen: vee-validate is 84KB minified / 23KB gzipped, so we should make a compelling case for bringing it in. Vuetify already provides some nice utilities for validation, what features of vee-validate do we need beyond that? Is it something we could add inexpensively without pulling in a third party library?

ping @jbeezley since I think he also has experience with vee-validate.

README.md
```diff
@@ -15,3 +16,60 @@
...
+Because these are vuetify components, your consumer application is responsible for creating the `v-app` container.  [Read more](https://vuetifyjs.com/en/layout/pre-defined#all-about-app).  See `src/main.js` for a more comprehensive example.
+
+### Components
+
+Public components live in `src/components` and should be imported by name.
+
+* Upload
+* Authentication
+
+### Event API
```
zachmullen: #11 is now merged to master and provides `login`, `logout`, and `register` events on the `girderRest` injection. So let's use that to avoid introducing this new bus.

README.md
```diff
@@ -19,10 +19,27 @@
...
+
+```bash
+# Vuetify
+vue add vuetify
+
+# Vee Validate
+yarn add vee-validate
+
+# FontAwesome
+yarn add @fortawesome/fontawesome-free -D
```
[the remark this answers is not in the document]
subdavis: Material Icons unfortunately lacks branded icons, so any company logos are missing.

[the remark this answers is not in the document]
zachmullen: I've been fairly disappointed with Material Icons. I'll open an issue to remove it in favor of FA, we can do it on a follow-up branch.

README.md
jbeezley: There are ways to [reduce the size](https://baianat.github.io/vee-validate/concepts/bundle-size.html#minimal-bundle).

I ended up using vee-validate for resonantgeo because I found the Vuetify's syntax to be excessively verbose, so I think it's something we could consider.  Here though, it's only being used to validate required components.  This can just be replaced by a prop (`:rules="[v => !!v || 'Item is required']"`) on the form element.

Another thing to note is that Vee-Validate [does not work](https://github.com/baianat/vee-validate/issues/1267) with vue-test-utils in synchronous mode.  For resonantgeo, I had to make all tests that touched the validation logic async.  It wasn't too bad, but it did require adding a lot of `await Vue.$nextTick()` statements.

For now, I would lean against using it until we have a more complicated use case.

src/App.vue
```diff
@@ -60,6 +53,10 @@ export default {
     if (this.girderRest.user) {
       this.fetchFolder();
     }
+    this.girderRest.$on(['login', 'register'], () => {
```
zachmullen: The `RestClient` actually emits the `login` event on successful registration, so that's the only one you need to listen to.

README.md
```diff
@@ -15,3 +16,47 @@
...
+# OR add the CDN link in index.html
+# <link href="https://use.fontawesome.com/releases/v5.0.13/css/all.css" rel="stylesheet">
+```
+
+To use components in your Vue.js app, the Vue instance must meet some requirements.  The basic requirements are:
+
+```javascript
+Vue.use(Vuetify);
```
zachmullen: This part is automatically handled as a side effect of importing components from index.js, so we can omit it from here.

README.md
zachmullen: Let's remove this whole section for now, it's probably premature to impose this as a documentation burden, and it's currently only documenting low-level events that are probably only useful to developers of this library. IMO the documentation for downstream-facing events should probably be auto-generated if possible.

README.md
```diff
@@ -15,3 +16,47 @@
...
+
+```bash
+# Vuetify
+vue add vuetify
+
+# FontAwesome
+yarn add @fortawesome/fontawesome-free -D
+# OR add the CDN link in index.html
+# <link href="https://use.fontawesome.com/releases/v5.0.13/css/all.css" rel="stylesheet">
+```
```
zachmullen: I think we can do away with this part; in #19 it just shows `yarn add @girder/components`, which will install vuetify, and the icon library could be invisible to downstreams depending on #20.

subdavis:
Notable changes in 4e6c812 :

* Modified package json to provide dependencies for pug and stylus to consumer
* Organized README with syntax descriptions for code blocks, clarified a bit on girderRest.

README.md
```diff
@@ -1,72 +1,107 @@
...
+>  import { Upload as GirderUpload } from '@girder/components/src/components';  // Good
+>  import GirderUpload from '@girder/components/src/components/Upload.vue'; // Unsafe -- may move in future
+> ```
+> Files and symbols that do not appear in an index.js should be considered private and it
+> is unsafe to use them in downstream projects since they are not part of the supported API surface.
 
-## For developers
-### Project setup
-    yarn
+### Event API
```
zachmullen: I think we should remove this section of the docs for now.

src/App.vue
```diff
@@ -1,26 +1,31 @@
...
-    v-text-field(v-model="password", label="Password", type="password")
-    v-alert(:value="error") {{ error }}
-    v-btn(@click="login") Login
+  girder-upload(v-if="folder" :dest="folder", :multiple="multiple")
+  v-layout(v-else justify-center)
+    v-flex(xs12 sm10 md8 lg6)
+      girder-auth(
+        :register="true"
+        :oauth="true"
+        forgot-password-link="http://localhost:8080/#?dialog=resetpassword")
```
zachmullen: Let's not include the origin here.

README.md
```diff
@@ -1,72 +1,107 @@
...
 
-If you're building your own downstream application, you can include individual components from
-the library. Either import the full UMD module:
+#### Girder RestClient
+
+Many components in this library will require a `girder.RestClient` named `girderRest` through provide/inject.  
+The client can be provided through any common ancestor.  For example:
+
+```javascript
+import * as girder from '@girder/components/src';
```
zachmullen: `import girder from '@girder/components';`

README.md
```diff
@@ -1,72 +1,107 @@
...
+# Project Setup
+yarn
+# Compile and hot-reload for development
+yarn serve
+# Build the library for production
+yarn build
+# Lint and fix source code
+yarn lint
+# Run unit tests
+yarn test:unit
```
zachmullen: Could we add a newline between each comment/command pair for readability?

subdavis:
Still working on a few things.

src/components/Authentication/Authentication.vue
```diff
@@ -9,7 +9,8 @@ v-flex.pa-2(xs-12)
           ref='loginForm'
           :oauth-providers='oauthProviders'
           :forgot-password-url='forgotPasswordUrl'
-          :forgot-password-route='forgotPasswordRoute')
+          :forgot-password-route='forgotPasswordRoute'
+          @forgotpassword='$emit("forgotpassword")')
```
subdavis: @zachmullen Are you OK with bubbling the event up to the parent like this so a consumer could listen for it?

subdavis: Note: App.vue is still passing a raw link.  You can now set routes, but since this app doesn't use a router yet, I figured that was the best option.  It will now correctly route you to the forgot password UI on the girder instance (in the demo)

zachmullen: Yep, that's a good practice.

README.md
```diff
@@ -1,72 +1,101 @@
...
+// Provide "girderRest" to the root
+new Vue({
+  provide: { girderRest },
+  ...
+}).$mount('#app');
+```
+
+#### Components
+
+If you're building your own downstream application, you can include individual components from the library. Because these are vuetify components, your consumer application is responsible for creating the `v-app` container.  [Read more](https://vuetifyjs.com/en/layout/pre-defined#all-about-app).  See `src/main.js` for a more comprehensive example.
```
zachmullen: I did an experiment to see what the Upload component looks like without a `v-app` containing it, and it worked fine. In the interest of keeping Vuetify as an internal detail whenever possible, what does this component look like if it's not inside `v-app`?

subdavis: Yes, the rendering changes for this app without v-app.  I'm utilizing some of the light/dark featuers.

> Vuetify requires the use of the v-app component. It should wrap your entire application and is the center point for much of the framework functionality. If for whatever reason you cannot use this element, you can mimic it from attributes and classes. Set the data-app attribute to true on the highest element available (not including body), and the application application--{light|dark} classes.

As far as I can tell, the mimicking doesn't work.  Without v-app, it looks like some CSS classes don't get loaded, so even though the DOM looks the same, v-app isn't present to cause the necessary assets to load.

Light/dark still works, but `color="primary"` is no longer functional.  I can use custom CSS to fix the things that are wrong, but that feels like a code smell.

subdavis: I'm pretty confident about classes not getting loaded without v-app.  I can force webpack to load the missing assets by putting an empty v-app back in as a leaf node anywhere in the DOM with `display:none;` and the styles come back.

zachmullen: Thanks for investigating, we shall keep this in the docs then.

src/components/Authentication/OAuth.vue
```diff
@@ -0,0 +1,89 @@
...
+  background-color: #283e4a !important;
+}
+
+.v-btn.box {
+  background-color: #0071f7 !important;
+}
+
+.v-btn.globus {
+  background-color: #335a95 !important;
+}
```
zachmullen: I'll try and add a PR to do stylint soon, but until then I will unfortunately act as a linter. The above can be rewritten as:

```stylus
.oauthIcon
  font-size 20px

 h4
  margin-left 0

 .v-btn.google
  background-color #3367d6 !important

 .v-btn.linkedin
  background-color #283e4a !important

 .v-btn.box 
  background-color #0071f7 !important

 .v-btn.globus
  background-color #335a95 !important
```

Basically, no curly braces, no colons, no semicolons. Nesting syntax is supported via `&`

```stylus
.v-btn
   &.box 
    background-color #0071f7 !important

  &.globus
    background-color #335a95 !important

...etc
```

src/components/Authentication/OAuth.vue
```diff
@@ -0,0 +1,89 @@
...
+};
+</script>
+
+<style lang="stylus" scoped>
+.oauthIcon {
+  font-size: 20px;
+}
+
+h4 {
+  margin-left: 0px;
```
zachmullen: Probably better to use the `.ml-0` class provided by Vuetify.

src/components/Authentication/OAuth.vue
```diff
@@ -0,0 +1,89 @@
...
+  methods: {
+    iconName(providerId) {
+      return `${iconMap[providerId].icon}`;
+    },
+  },
+};
+</script>
+
+<style lang="stylus" scoped>
+.oauthIcon {
```
zachmullen: I think this class is actually unused.

package.json
```diff
@@ -29,9 +33,6 @@
     "babel-core": "7.0.0-bridge.0",
     "babel-jest": "^23.6.0",
     "pug-lint-vue": "^0.1.3",
-    "stylus": "^0.54.5",
-    "stylus-loader": "^3.0.2",
-    "vue-cli-plugin-pug": "^1.0.7",
```
subdavis: For the sake of flexibility, I replaced this with pug and pug-plain-loader and move them from devDependencies to dependencies, since the consumer will need a loader to compile our pug. This allows them to use vue-cli-3, vue2, or whatever else with their own webpack configuration.

vue-cli already supports this out-of-the-box, so all you have to do is have pug installed.

https://github.com/vuejs/vue-cli/blob/f8e97838dd25e37433ebbdb92b748c48b89f7c2a/packages/%40vue/cli-service/lib/config/base.js#L131

zachmullen: On #29, I intentionally replaced `pug-loader` and `pug-plain-loader` with `vue-cli-plugin-pug`. The rationale there is that the loader chain used by vue-cli-service to support pug is actually an internal detail, and using the official Vue plugin that supports pug in components means we don't have to know that implementation detail. While it may be obvious that it's pug and pug-loader, upstream vue-cli-service could add additonal loaders or loader configuration that requires other packages, and by virtue of using the plugin, we don't have to worry about that at all.

Regarding whether things belong in dependencies of devDependencies, I'm actually not sure what the best solution here, I'd like to hear @jbeezley 's thoughts on this as well. The way I had separated it, and the way @jbeezley chose to separate it on resonantgeo, was to put runtime dependencies in `dependencies` and build-time dependencies in `devDependencies`. It kind of depends in what context people are installing the package whether that makes sense -- if they are just installing it to use the prebuilt UMD module, they don't need anything other than perhaps a peer dependency on `vue`. If they want to build a downstream app using the source files in this package, they'll need all our runtime dependencies, but will likely be providing their own build toolchain, which I think is why we had it setup like that.

I'm not sure what the correct answer is, but I think it makes sense that all of our build time dependencies should be in the same place, and all of our runtime dependencies should be in the same place.

src/components/Authentication/Register.vue
```diff
@@ -0,0 +1,133 @@
...
+  watch: {
+    password() {
+      this.validatePasswordsMatch();
+    },
+    retypePassword() {
+      this.validatePasswordsMatch();
+    },
+  },
+  methods: {
+    listify: v => (v ? [v] : []),
```
zachmullen: This method is unused (looks like it may be leftover copypasta from girder-vue)

src/rest.js
```diff
@@ -85,6 +89,13 @@ export default class RestClient extends Vue {
     return this.user;
   }
 
+  async fetchOauth(redirect = window.location.href, list = true) {
+    const resp = await this.get('oauth/provider', {
```
zachmullen: For Girder servers that do not have the OAuth plugin enabled, this will return some sort of 4xx response, so we should wrap this line in a try/catch, and if we detect an exception we can `return []`.

package.json
subdavis: Gotcha - sorry for the assumption here, and thanks for the thorough explanation.

I went through the exercise of creating a downsteam using this as source and these were the dependencies not already included by virtue of downstream necessarily being a Vue.js app.

src/components/Authentication/Authentication.vue
```diff
@@ -0,0 +1,94 @@
...
+          name: provider.name,
+          url: provider.url,
+        });
+      });
+    }
+  },
+};
+</script>
+
+<style lang="stylus">
```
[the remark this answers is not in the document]
subdavis: I intentionally made these styles without `scoped` because they're used in all the children -- OAuth, Register, Login, etc.

I didn't see a policy for duplicate CSS in this project, so I named a class after this module to manage the scope.  I saw this policy implemented in [resonantgeo](https://github.com/Kitware/resonantgeo/blob/b5665b998de0de6731546aa2302d9d182f16d0db/src/components/AppToolbar.vue#L45).

We can totally duplicate these styles into the other vue components, but it may be more difficult to keep styles consistent if you have to check all children when styles are changed.  I'd like to get @jbeezley's thoughts here as well.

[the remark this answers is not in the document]
jbeezley: That's done in resonantgeo because it alters the CSS of the child component `.v-toolbar__content` for which there is no other way.

I haven't though about *sharing* CSS between components.  This would work, but would implicitly assume that the child components are not used with out the parent.  A stackoverflow answer mentions [css modules](https://vue-loader.vuejs.org/guide/css-modules.html) as a possible strategy, but I don't really know anything about them.  Another possibility is a `.styl` file that is imported by all of the components as
```html
<style lang="stylus" src="./common.styl" scoped></style>
```

package.json
jbeezley: There really isn't a "correct" answer.  The philosophy of the webpack maintainers is that loaders and webpack configs are an application concern, so they should go in `devDependencies`, but you can come up with arguments for why that is bad.

I think since we are publishing a UMD bundle, we should optimize for it.  That means keeping them in `devDependencies` so that we minimize the packages that are installed.  If someone is importing the components directly, that is an advanced use case and they are on their own in terms of importing loaders.

src/components/Authentication/Authentication.vue
```diff
@@ -0,0 +1,94 @@
...
+    padding-bottom 2px
+</style>
+
+<style lang="stylus" scoped>
+.v-alert.error
+  margin 0px
+  padding 10px 16px
+.v-alert__icon
+  margin-right 8px
+</style>
```
jbeezley: Are either of these styles actually doing anything?  They seem to reference elements in child components.

src/components/Authentication/Authentication.vue
```diff
@@ -0,0 +1,94 @@
+<template lang="pug">
+v-flex.pa-2.girder-authentication-component(xs-12)
```
zachmullen: We should remove this outer element. To make this reusable in a variety of contexts, we'll let callers specify what sort of layout or container it should be in. To that end, I also think we should either make the outer `v-card(flat)`, or remove it altogether so that downstreams can decide for themselves whether they want box shadow / border radius around this element. Another alternative would be to add `flat` as a prop of this component.

As a general rule, we shouldn't add padding or margin to the top level elements of a reusable component, we should only use it to separate elements inside them. We should think of these components as just the "inside" of some containing element, and we should add the containing elements inside our `App.vue` example.

src/App.vue
```diff
@@ -59,6 +55,10 @@ export default {
     if (this.girderRest.user) {
       this.fetchFolder();
     }
+    this.girderRest.$on('login', () => {
+      this.girderRest.fetchUser();
```
zachmullen: No need to call `fetchUser` here, if you have received the 'login' event, the user will already be set on the `girderRest` instance.

src/rest.js
```diff
@@ -85,6 +89,13 @@ export default class RestClient extends Vue {
     return this.user;
   }
 
+  async fetchOauth(redirect = window.location.href, list = true) {
```
zachmullen: One design decision we (@jbeezley @danlamanna and I) made regarding the `RestClient` class is that we don't want to populate it with wrapper methods for "ordinary" REST requests. Things like login/logout/register/fetchUser are provided because they actually change the state of the `RestClient` instance itself and emit special events. I could also see exceptions made for certain kinds of particularly nontrivial requests, though to be honest I have not really encountered any other than uploading files. Since this call is trivial, let's remove this method and just make the call directly from the component.

src/components/Authentication/Authentication.vue
```diff
@@ -0,0 +1,94 @@
...
+import Login from './Login.vue';
+import Registration from './Register.vue';
+
+export default {
+  inject: ['girderRest'],
+  props: {
+    register: {
+      type: Boolean,
+      default: false,
+      note: 'Enable registration from this component?',
```
zachmullen: I like documenting the props, but unless there is some kind of runtime support for this field (reading the Vue docs I couldn't find anything about "note" on props), we should instead document in comments, which can be stripped out during bundling.

We plan to use vue-styleguidist to document these, so follow [their conventions](https://github.com/vue-styleguidist/vue-styleguidist/blob/master/docs/Documenting.md). Unfortunately there's an upstream issue preventing us from doing this right now, which is tracked by #14, but once that's resolved we should just be able to flip a switch.

src/components/Authentication/Authentication.vue
```diff
@@ -0,0 +1,94 @@
...
+    if (this.oauth) {
+      const resp = await this.girderRest.fetchOauth();
+      this.oauthProviders = [];
+      resp.data.forEach((provider) => {
+        this.oauthProviders.push({
+          id: provider.id,
+          name: provider.name,
+          url: provider.url,
+        });
+      });
```
zachmullen: We can replace the above with

```javascript
if (this.oauth) {
  try {
    this.oauthProviders = (await this.girderRest.get('oauth/provider', {
      params: {
        redirect: window.location.href,
        list: true,
      },
    })).data;
  } catch (e) { /* use default=[] */ }
}
```

src/components/Authentication/Authentication.vue
zachmullen: I completely removed both of the style blocks from the parent component, and moved the `no-decorate` rule to the OAuth component, and everything looked fine, so my vote is to do that.

src/components/Authentication/OAuth.vue
```diff
@@ -0,0 +1,73 @@
+<template lang="pug">
+.oauth-widget(v-if="providers.length")
```
zachmullen: We don't use the `oauth-widget` class anywhere, so it can be removed. Also, we should move this `v-if` up to the parent components, I think it may generally be a code smell to have the top level element contain `v-if`. So, we can get rid of this element altogether.

src/rest.js
```diff
@@ -35,6 +35,10 @@ export default class RestClient extends Vue {
     }));
   }
 
+  get url() {
+    return new URL(this.apiRoot);
```
zachmullen: This actually throws an exception in production, where the default `apiRoot` is "/api/v1". I think we can get rid of this method and let callers do this in the uncommon case where they need to parse the URL (YAGNI).

src/components/Authentication/OAuth.vue
```diff
@@ -0,0 +1,73 @@
...
+  v-container
+    h4.ml-0 Or {{verb}} with OAuth
```
zachmullen: This `ml-0` seems to have no effect in my local experiment.

src/components/Authentication/OAuth.vue
```diff
@@ -0,0 +1,73 @@
...
+    a.no-decorate(v-for="provider in providers",
+        :key="provider.id",
+        :href="provider.url")
+      v-btn(:dark="iconMap[provider.id].dark", :color="provider.id")
+        v-icon.aligned {{ $vuetify.icons[iconMap[provider.id].icon] }}
```
zachmullen: Is the `aligned` class something from vuetify? It doesn't seem to have any effect for me locally.

src/components/Authentication/OAuth.vue
```diff
@@ -0,0 +1,73 @@
+<template lang="pug">
+.oauth-widget(v-if="providers.length")
+  v-container
+    h4.ml-0 Or {{verb}} with OAuth
+    a.no-decorate(v-for="provider in providers",
```
zachmullen: `v-btn` has an `href` property, so we can get rid of this element and the `no-decorate` class.

```javascript
    v-btn(v-for="provider in providers",
        :key="provider.id",
        :href="provider.url",
        :dark="iconMap[provider.id].dark", :color="provider.id")
      v-icon.aligned {{ $vuetify.icons[iconMap[provider.id].icon] }}
      .ml-2 {{ provider.name }}
```

src/components/Authentication/OAuth.vue
```diff
@@ -0,0 +1,73 @@
...
+        :key="provider.id",
+        :href="provider.url")
+      v-btn(:dark="iconMap[provider.id].dark", :color="provider.id")
+        v-icon.aligned {{ $vuetify.icons[iconMap[provider.id].icon] }}
+        span &nbsp; {{ provider.name }}
```
zachmullen: Rather than using nbsp for spacing, we should use Vuetify's support for icons in buttons:

```pug
v-icon(left) {{ $vuetify.icons[iconMap[provider.id].icon] }}
| {{ provider.name }}
```

src/components/Authentication/OAuth.vue
subdavis: I prefer these over plain divs because it makes navigating the inspector easier.  You can see by class exactly where components start.  This is especially handy to me because vuetify components seems to add 3-5 layers of wrapping, so the tree is kind of noisy.

Vuetify also follows this practice, wrapping each component in a div with class `v-component-name`

zachmullen: Fair enough, we can just move the class down to the `v-container`.

subdavis:
@zachmullen @jbeezley 

I think I made adjustments for all comments except for the ones about scoped styles.  After pruning the styles, here's a before and after for the un-scoped styles that need to be shared between Auth and its children.  Most of the changes can be seen in the vertical alignment of form fields, icons, and buttons on the left.

My motivation for keeping these together in the parent, in addition to them applying in 2 or more places,  is that they all impact the unified Authentication view.  The changes to button margins, icon margins, etc. all result in alignment of the different parts of auth.  If I were to re-use the OAuth or Login components on their own, these styles would serve no purpose because there might be different things to align with.

screenshot_20181001_142948
screenshot_20181001_143059

src/components/Authentication/OAuth.vue
```diff
@@ -0,0 +1,58 @@
+<template lang="pug">
+v-container.oauth-widget
+  h4 Or {{ verb }} with OAuth
+  v-btn.no-decorate(v-for="provider in providers",
```
zachmullen: Now that this is using `v-btn`, `no-decorate` is no longer necessary.

src/components/Authentication/OAuth.vue
```diff
@@ -0,0 +1,58 @@
...
+
+<style lang="stylus" scoped>
+.no-decorate
+  text-decoration none
+
+.v-btn.google
+  background-color #3367d6 !important
+
+.v-btn.linkedin
+  background-color #283e4a !importan
```
zachmullen: typo in `!important`

src/components/Authentication/OAuth.vue
```diff
@@ -0,0 +1,58 @@
...
+
+.v-btn.box
+  background-color #0071f7 !important
+
+.v-btn.globus
+  background-color #335a95 !important
```
zachmullen: I think we should nest the above rules for the sake of DRY:

```stylus
.v-btn
  &.globus
    background-color #335a95 !important

  &.box
    background-color #0071f7 !important

  // etc
```

That way if for some reason we use something other than `.v-btn` as the containing element we would just have to change it in one line rather than several.

src/components/Authentication/Register.vue
```diff
@@ -0,0 +1,116 @@
+<template lang="pug">
+.register-widget
+  v-alert(
+      dismissible,
```
zachmullen: Should this be dismissable? Out of my depth, paging @jtomeck

zachmullen:
Part of the weird alignment is that Vuetify adds 24px of padding to `v-container`, but only 16px if the `v-container` is inside of a `v-form`. Simply switching the `v-container` to be outside the `v-form` rather than inside fixed some of the alignment straight away.

After that, the only misaligned thing was the buttons with the icons and text, but that's only because buttons by default have a bit of margin on all sides. A quick solution to that is just adding `.ml-0` to the two submit buttons. For the OAuth buttons, I added `.ml-0.mr-3` which I think looks nice, and then we don't need any shared styles.

src/components/Authentication/Login.vue
```diff
@@ -0,0 +1,110 @@
...
+          :rules="nonEmptyRules",
+          prepend-icon="$vuetify.icons.user",
+          type="text")
+      v-text-field(
+          v-model="password",
+          type="password",
+          label="Password",
+          :rules="nonEmptyRules",
+          prepend-icon="$vuetify.icons.lock")
+      .submit-area
```
zachmullen: We can get rid of the style block if we change this to a `v-layout(row)` and then add `v-spacer` between the two buttons.

src/components/Authentication/OAuth.vue
```diff
@@ -0,0 +1,53 @@
...
+      iconMap,
+    };
+  },
+};
+</script>
+
+<style lang="stylus" scoped>
+.v-btn
+  &.google
+    background-color #3367d6 !important
```
zachmullen: Sorry for the run-around on this particular section of code -- I re-read it and realized we are doing something that may be relying on undefined behavior. You are passing `:color="provider.id"` to the button, which does add this class, but that may be an implementation detail. I think it would be better to remove this style block altogether, add the colors as a new field into the `iconMap`, and pass `:color="iconMap[provider.id].color"`.

Aside for general proselytizing: @jeffbaumes has ingrained the idea in me that we should have as little code in `style` blocks as possible, as they (often) represent deviations from the norms of our styling framework.

src/components/Authentication/Authentication.vue
```diff
@@ -0,0 +1,69 @@
...
+            :forgot-password-route="forgotPasswordRoute",
+            @forgotpassword="$emit('forgotpassword')")
+      v-tab-item(key="registration-box", v-if="register")
+        registration(ref="registerForm",
+            :oauth-providers="oauthProviders")
+</template>
+
+<script>
+import Login from './Login.vue';
+import Registration from './Register.vue';
```
zachmullen: These are internal details, but for the sake of propagating best practices, we should use:

```javascript
import GirderLogin from './Login.vue';
import GirderRegistration from './Register.vue';

...
components: {
  GirderLogin,
  GirderRegistration,
},
```

And same for the OAuth component usage.

Thanks @jeffbaumes for teaching me this practice, which is [recommended by Vue](https://vuejs.org/v2/style-guide/#Multi-word-component-names-essential).

src/components/Authentication/Login.vue
```diff
@@ -0,0 +1,104 @@
...
+        v-btn.ml-0(type="submit",
+            color="primary",
+            :disabled="inProgress",
+            :loading="inProgress")
+          v-icon.aligned(left) $vuetify.icons.login
+          | Login
+        v-spacer
+        v-btn(flat, color="primary", @click="forgotPasswordAction") Forgot Password?
+  v-divider(v-if="oauthProviders.length")
+  oauth(ref="oauth", v-if="oauthProviders.length", :providers="oauthProviders")
```
zachmullen: This `ref` is not used

src/components/Authentication/Authentication.vue
```diff
@@ -0,0 +1,69 @@
...
+      v-tab(key="login") Log In
+      v-tab(key="registration", v-if="register") Register
+      v-tab-item(key="login-box")
+        login(ref="loginForm",
+            :oauth-providers="oauthProviders",
+            :forgot-password-url="forgotPasswordUrl",
+            :forgot-password-route="forgotPasswordRoute",
+            @forgotpassword="$emit('forgotpassword')")
+      v-tab-item(key="registration-box", v-if="register")
+        registration(ref="registerForm",
```
zachmullen: This `ref` and the `loginForm` ref are unused

src/components/Authentication/Register.vue
```diff
@@ -0,0 +1,116 @@
...
+      v-text-field(
+          v-model="retypePassword",
+          type="password",
+          label="Retype password",
+          :rules="retypeMustMatchPasswordRules")
+      v-btn.ml-0(type="submit",
+          color="primary",
+          :loading="inProgress") Register
+  v-divider(v-if="oauthProviders.length")
+  oauth(ref="oauth",
```
zachmullen: This `ref` is not used

src/components/Authentication/Login.vue
```diff
@@ -0,0 +1,104 @@
...
+          type="password",
+          label="Password",
+          :rules="nonEmptyRules",
+          prepend-icon="$vuetify.icons.lock")
+      v-layout(row)
+        v-btn.ml-0(type="submit",
+            color="primary",
+            :disabled="inProgress",
+            :loading="inProgress")
+          v-icon.aligned(left) $vuetify.icons.login
```
zachmullen: Is the `aligned` class something from Vuetify? It doesn't seem to have any effect for me, but it's in a handful of places in these components.

subdavis: No, I added this because I didn't like the vertical alignment of icons in buttons.  They looked out of line with the text to me.  The actual styles are gone now. These were lingering and needed to be removed.

src/components/Authentication/Login.vue
```diff
@@ -0,0 +1,104 @@
...
+          | Login
+        v-spacer
+        v-btn(flat, color="primary", @click="forgotPasswordAction") Forgot Password?
```
zachmullen: I'd suggest changing this to

```pug
v-btn(
    flat, color="primary", :to="forgotPasswordRoute", :href="forgotPasswordUrl",
    @click="$emit('forgotPassword')") Forgot Password?
```

And removing the `forgotPasswordAction()` method. The nice thing about this way would be that if a URL or route was passed, the button becomes a link and users can open it in a new tab, copy the URL with right click, etc.

src/components/Authentication/Authentication.vue
```diff
@@ -0,0 +1,67 @@
...
+    },
+    /* A full URL to be used as an anchor href to an external page.
+     * Overrides forgotPasswordRoute.
+     */
+    forgotPasswordUrl: {
+      type: String,
+    },
+    /* A vue-router route path.  Must use object form. */
+    forgotPasswordRoute: {
+      type: Object,
```
zachmullen: `type: [Object, String]`, since a route can be expressed as either type and the underlying components support either.

tests/unit/Authentication.spec.js
```diff
@@ -0,0 +1,61 @@
...
+        register: true,
+        oauth: false,
+        forgetPasswordLink: 'https://example.com/',
+      },
+      provide: { girderRest: new RestClient() },
+    });
+    expect(wrapper.contains(Register)).toBe(true);
+  });
+
+  it('fetches Oauth Providers if Oauth is enabled', (done) => {
```
zachmullen: These test cases can actually accept `async` functions so we don't have to do the callback thing. Then the bottom part can be replaced with 

```javascript
await wrapper.vm.$nextTick();
expect(wrapper.vm.oauthProviders.length).toBe(1);
```

tests/unit/Authentication.spec.js
```diff
@@ -0,0 +1,61 @@
...
+    const get = () => Promise.resolve({
```
zachmullen: We've created a nice pattern for mocking REST interaction that can be seen in uploadUtil.spec.js (among others).

We use [axios-mock-adapter](https://github.com/girder/girder_web_components/blob/fc44f25ce350bdf89647b310c5c77564c5ae1a02/tests/unit/uploadUtil.spec.js#L8) to mock the XHRs coming out of the RestClient, [like this example](https://github.com/girder/girder_web_components/blob/fc44f25ce350bdf89647b310c5c77564c5ae1a02/tests/unit/uploadUtil.spec.js#L55). For this, it'd look like:

```javascript
mock.onGet(/oauth\/provider/).replyOnce(200, [{
  id: 'foo',
  name: 'bar',
  url: 'https://example.com/',
}]);
```

The nice thing about using this way is that it makes it quite easy to maintain the test mocking even as components become more complex and make more requests, and `axios-mock-adapter` makes it very straightforward to mock rejected XHRs of all types (network failure, HTTP error responses, or timeouts).

src/components/Authentication/Register.vue
```diff
@@ -0,0 +1,115 @@
...
+      retypeMustMatchPasswordRules,
+    };
+  },
+  methods: {
+    async register() {
+      if (!this.$refs.register.validate()) {
+        return;
+      }
+      this.inProgress = true;
+      this.alerts.errors.length = 0;
```
zachmullen: I didn't realize this actually works as a way to empty an Array, you taught me something new! (I would've expected `length` to be a read only property.)

To follow the Principle of Least Astonishment, it's probably best to instead do `this.alerts.errors = [];`.

src/components/Authentication/Login.vue
```diff
@@ -0,0 +1,97 @@
...
+        v => !!v || 'Item is required',
+      ],
+    };
+  },
+  methods: {
+    async login() {
+      if (!this.$refs.login.validate()) {
+        return;
+      }
+      this.alerts.errors.length = 0;
```
zachmullen: (see my comment below)

tests/unit/Authentication.spec.js
subdavis: There are some things I likely don't understand about what this mocker is doing.

With my solution that directly uses `Promise.resolve()`, the tests pass with a single call to `await nextTick()`

If I replace that with `mock.onGet()`, I need 3 or more calls.  I checked and there's a 6-10 millisecond delay between the call and response for both.  This also shouldn't make a difference to the vue instance.

I tried experimenting with returning a promise from mock.onGet, but this still required exactly 3 calls to nextTick.

zachmullen: That's weird. I hadn't tried this on an actual Vue component, so I haven't run into this myself. I will take a look at it today and see if I can sort it out, if so I'll just append a commit onto this branch and we can merge, otherwise we can use your previous solution for now.

jbeezley: It might have to do with `mounted` being an async function, which interferes somehow with how vue-test-utils tries to make components synchronous.  There is a lot of traffic in the Vue issue tracker about this kind of thing.  The [vue-test-utils docs](https://vue-test-utils.vuejs.org/guides/#testing-asynchronous-behavior) mention using [flush-promises](https://github.com/kentor/flush-promises).  That seems to avoid the need for multiple `await`'s in this test.

As a side note on the oauth list, I am a fan of using [vue-async-computed](https://github.com/foxbenjaminfox/vue-async-computed) for async properties.  It might be too much to bring in for this one case, but I suspect it is a pattern we will need again.

jbeezley: Agreed

subdavis: Test has been commented out with a reference to this thread.

PR #19, opened 2018-09-17
Add basic instructions for users

zachmullen:


zachmullen:
@jbeezley @brianhelba please take a look at 10eeebf as it publishes a policy about what is and is not part of our API surface for this library.

zachmullen:
@subdavis could you re-add your review if you like my changes.

PR #22, opened 2018-09-20
Switch to Font Awesome

zachmullen:
Refs #17


src/components/Upload.vue
```diff
@@ -30,17 +30,18 @@ v-card.upload-wrapper(flat)
 
   slot(name="files")
     v-list(v-show="files.length", dense)
-      div(v-for="(file, i) in files", :key="file.file.name")
+      .file-tile(v-for="(file, i) in files", :key="file.file.name",
```
zachmullen: I moved this to fix a bug where the v-dividers from completed uploads persisted after the upload was completed, and they stacked up on top of each other.

jeffbaumes: Good catch

PR #24, opened 2018-09-20
Lint pug templates in Vue components

zachmullen:
Unfortunately, the vue-cli-service tooling provides no support for linting pug templates, and has no plans to at the moment, though [open issues exist](https://github.com/vuejs/eslint-plugin-vue/issues/310), and I imagine it will be added at some point in the future.

I think we should add some third party tooling to support this for the moment. Yes, this adds an annoying amount of dev dependencies and departs slightly from the world of pure first-class vue-supported tools, however I think it's important enough as we gain momentum writing components to justify this.

It's clear that code review time is one of our most limited resources, and we know that using humans as linters in code review is an anti-pattern / bad practice. So with that in mind, I think we should merge this, and open an issue to remove most of this code in the future once pug support gets added to vue tooling.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,7 @@ jobs:
           paths:
             - ~/.cache/yarn
       - run: yarn lint --no-fix
+      - run: yarn lint:pug
       - run: yarn test:unit --coverage --coverageReporters html
       - store_artifacts:
           path: coverage/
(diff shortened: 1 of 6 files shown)


zachmullen:
Sorry for the massive yarn.lock changes, there was a conflict in it, and my answer to yarn.lock conflicts is `rm yarn.lock ; yarn`.

PR #28, opened 2018-09-22
Switch to Material Design Icons

zachmullen:
In addition to just changing the icon set, this establishes a new practice of referencing semantic icon variables rather than encoding literal icon classes in components. This will make it easy to switch icon libraries or change up icons based on semantics in the future.

**Before:**

`v-icon mdi-close`

**New practice:**

`v-icon $vuetify.icons.close`

If you need to add an entry to this variable map, it lives in `src/components/vuetifyConfig.js`, which is a private module. It uses the 'mdi' preset, and we will add our own extensions / overrides into the `icons` property as needed.


zachmullen:
@jtomeck @subdavis PTAL.

FWIW, I also like the way this set looks better than FontAwesome :)

zachmullen:
Thanks for approval @jbeezley 

I will merge this, but I wanted to mention something that I grappled with while developing this, which we will need to address somehow in the future:

Importing our components installs the Vuetify plugin with some custom configuration options. There's something problematic in this, which is that installing plugins into the global `Vue` can *only be done once* -- any future calls of the form `Vue.use(Vuetify, ...)` will simply be a no-op. A more desirable behavior would be that any configuration options passed as the second argument would be merged into the existing configuration, but that's not how it works, and probably for good reason. So, downstreams of this library who are also using Vuetify themselves (probably most of them) will be forced to accept Girder's Vuetify configuration as is, or if they need to alter it, they must do so dynamically by setting properties on the `$vuetify` property of their Vue *instance*, which is a little ugly.

One solution I considered was to require the downstreams to install the Vuetify plugin themselves and include Girder's configuration, which is nice in that it avoids this problem and removes one import side effect, but then the fact that we are using Vuetify is no longer an internal detail, which is a desirable property of the current architecture.

If we can come up with a solution that works robustly with Vuetify-based downstreams without necessarily leaking our dependence on Vuetify to everyone else, that would be ideal. At the end of the day, this could just be a documentation issue.

jbeezley:
Yeah, that's one of the reasons why I decided to make resonantgeo a Vue plugin.  Using it as a plugin can initialize all of the deps for you.  This leaves open the option to import the component directly and avoid the automatic initialization of vuetify.

jbeezley:
I think it would also work for downstreams to initialize vuetify before `Vue.use`-ing girder to override the config.

PR #30, opened 2018-09-28
Create a Vue plugin for this library

zachmullen:
@jbeezley PTAL, this follows our discussion from this morning. The first commit here represents the code changes; if you approve of those, I'll add a second commit for the documentation that is relevant to downstreams.

@subdavis PTAL

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,9 +1,23 @@
+import Vuetify from 'vuetify';
 import * as components from './components';
 import RestClient from './rest';
 import * as utils from './utils';
+
+/**
+ * This installs the Vue plugin for this library. This is exported in the top level module,
+ * making it a Vue plugin.
+ * @param Vue the Vue prototype to install the plugin into.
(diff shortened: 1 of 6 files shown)


subdavis:
Awesome, I was wondering about this as well.  LGTM.

src/index.js
```diff
@@ -1,8 +1,10 @@
 import * as components from './components';
+import GirderVue from './GirderVue';
 import RestClient from './rest';
 import * as utils from './utils';
 
 export {
+  GirderVue,
```
zachmullen: If we import this and export it as an `install` function, people could install the plugin just by:

```javascript
import girder from '@girder/components';
Vue.use(girder);
```

Is there any reason not to do it that way?

jbeezley: No.  I was going to say that is the way it is usually done.  I didn't even know you could pass the install function until I looked at the source.

PR #34, opened 2018-10-02
Add working nextTick helper and test utils module

zachmullen:
@subdavis @jbeezley this is the fix for the weirdness in the test we saw earlier today. That one liner in `utils.js` is functionally equivalent to the `flush-promises` package.

### DIFF
--- a/tests/unit/Authentication.spec.js
+++ b/tests/unit/Authentication.spec.js
@@ -5,6 +5,7 @@ import RestClient from '@/rest';
 import Authentication from '@/components/Authentication/Authentication.vue';
 import Register from '@/components/Authentication/Register.vue';
 import Oauth from '@/components/Authentication/OAuth.vue';
+import { flushPromises } from './utils';
 
 const localVue = createLocalVue();
 localVue.use(Vuetify);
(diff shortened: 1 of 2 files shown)


jbeezley:
I know this is very simple and works, but I'm conflicted about it.

1. The name `nextTick` is confusing because it overloads what vue means by nextTick, which is not necessarily the same thing as `setImmediate`.  (The implementation is [complicated](https://github.com/vuejs/vue/blob/dev/src/core/util/next-tick.js))
2. It uses `setImmediate` which works in node, but not in most browsers and is not in any standards proposal.  If we ever get around to testing on a browser, we will have to polyfill.

I'm somewhat inclined just to recommend the flush-promises library because that's what the vue-test-utils docs advise.

zachmullen:
I've revised the code, changing the name from `nextTick` to `flushPromises` to avoid confusion, and switching from `setImmediate` to `setTimeout` for portability.

I still think we should just do the implementation of flushPromises ourselves since it's literally just `new Promise(r => setTimeout(r))`. I just have a moral objection to bringing in a third party library for something so simple and built-in. I would argue that if something changes and this becomes insufficient, and the upstream implementation of `flush-promises` changes in some way that accommodates us in the future, we can easily switch by just plugging it into the method in place of what's written in `utils` now.

PR #36, opened 2018-10-05
Data Browser

### LINKED ISSUE #31: Create File Browser Component
Create a file browser component consistent with Jared's designs attached.

Questions:

* What are the possibilities for the breadcrumb root?  Anything other than a user's filespace?
* What does it look like when you create a new folder / new item?  Does @jtomeck have designs for those?
  * If this is up to me, should we go with a popup like the current girder UI?  If so, we may want to discuss how those look and feel.

Thanks!

[girder_ExportedFlow.zip](https://github.com/girder/girder_web_components/files/2435244/girder_ExportedFlow.zip)

subdavis:
fixes #31


subdavis:
I'm having a bit of trouble with balancing features with code efficiency and complexity, so I wanted to ask a couple questions for what features are important:

The main model is a `location` prop that determines where the browser's initial position is.  It has fields `id, type` which correspond to a girder Id and one of `collection, user, folder`. 

What is a reasonable set of features for the consumer to have?  I think it should be able to:
* Specify the initial location for the browser
* Maintain the model for location, such that:
  * it can intercept or block or block location changes
  * it can change the browser location at any time.
The consumer will likely also want to initialize you to a particular location, so `FileBrowser` will need to build the breadcrumb backwards with iterative api calls.  The consumer may also want to forcefully change your location in the browser based on events in other parts of the app, so the component should be able to do this any time the location changes.

Based on these features, and the constraints I set for myself:

* Don't use watched properties,
* Don't use mounted or created for anything other than setting up event listeners
* Use async computed for any properties that rely on server data (credit @jbeezley - I really like this library)
* Don't make unnecessary API queries.

I'm running into an issue with re-computation of the breadcrumb.  I could fairly easily do what I want by caching all the folder paths I've already seen, and only issuing an API call when I arrive at a parent folder I don't already have the details for.  This is, however, a little messy.

src/components/Authentication/OAuth.vue
```diff
@@ -17,7 +17,7 @@ const iconMap = {
   linkedin: { dark: true, icon: 'linkedin', color: '#283e4a' },
   bitbucket: { dark: false, icon: 'bitbucket' },
   box: { dark: true, icon: 'box_com', color: '#0071f7' },
-  globus: { dark: true, icon: 'globus', color: '#335a95' },
+  globus: { dark: true, icon: 'globe', color: '#335a95' },
```
zachmullen: We should isolate this in its own PR since it's unrelated to the topic.

subdavis: I'm using the `globe` icon in the file browser, so I changed its name to something generic.   Should I create a duplicate icon and then remove the duplicate in another PR?

zachmullen: Ah, I see. I'd actually like to keep the `globus` icon identifier separate from `globe`, even though they would point to the same thing right now. One is a brand and one is generic, so it makes sense to have two different semantic identifiers for them.

src/components/FileBrowser/FileBrowser.vue
```diff
@@ -0,0 +1,252 @@
...
+        v-breadcrumbs.pl-0
+          v-icon.mdi-24px(slot='divider', color='accent') {{ $vuetify.icons.chevron }}
+          v-breadcrumbs-item.selectable(@click.native="$emit('update:location', breadcrumb.root)")
+            v-icon.mdi-24px(color='accent') {{ $vuetify.icons.globe }}
+            | &nbsp; {{ login }}
+          v-breadcrumbs-item(v-for="item in breadcrumb.path", :key="`${item.id}.crumb`", @click.native="$emit('update:location', item)") {{ item.name }}
+      th(width="1%").pr-0
+        v-btn(flat, small, v-if="newFolderEnabled")
+          v-icon.mdi-24px.mr-1(left, color="accent") {{  $vuetify.icons.folderNew }}
+          | NEW FOLDER
```
jeffbaumes: I believe v-btn will automatically capitalize text in buttons. I'd prefer we keep them "Sentence case" in markup. The upcoming iteration of material design seems to remove the idea of uppercasing everything so it is a forward-compatible practice.

src/components/FileBrowser/FileBrowser.vue
```diff
@@ -0,0 +1,252 @@
+<script>
```
jeffbaumes: Interesting to do JS first, then template, then CSS. I don't mind this, but we should standardize on an ordering in Vue files for consistency.

subdavis: I picked this up at our local DC Vue meetup group.

The justification is that you always look at (html, script) or (html, css) but never (script, css), so it makes more sense to have your adjacency like this.

jeffbaumes:
The breadcrumb is always annoying to think about, and there are odd edge cases (e.g. where the user does not have access to every folder up the chain to a user/collection).

> Don't make unnecessary API queries.

I personally wouldn't worry about caching and would relax this constraint. Code clarity probably trumps efficiency at this point in the development of these components.

src/components/FileBrowser/FileBrowser.vue
matthewma7: I have been putting template, script, then style. but script, template, style actually feels better based on the frequency of the use. 
I learned a few style from the official style guide, https://vuejs.org/v2/style-guide/#Single-file-component-top-level-element-order-recommended

zachmullen: I'll make a PR to add `plugin:vue/recommended` to our eslint extensions.

zachmullen: #37 does this, but doesn't appear to include enforcement of the top level tag ordering... will look into this further after lunch.

src/App.vue
```diff
@@ -1,18 +1,26 @@
...
+  v-dialog(v-model="loginDialog", width="50%", persistent)
+    girder-auth(
+        :register="true",
+        :oauth="true",
+        :forgot-password-url="forgotPasswordUrl")
+  v-dialog(v-model="uploader", width="70%")
+    girder-upload(v-if="uploader",
+        :dest="uploadDest",
+        multiple="multiple",
+        @done="refresh++")
```
subdavis: I don't especially like this, but I also don't see a way around it.  In order to force computed properties to re-compute without any of their dependencies having changed where the trigger is an event from Upload.vue, this seems like the best way to do it.  

On the plus side, if the client wanted to implement some kind of polling loop to keep the file browser up to date, this would make it trivial.

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,294 @@
...
+      if (newval.id !== oldval.id) {
+        this.pagination.page = 1;
+      }
+    },
+    selected(newval) {
+      this.$emit('update:selected', newval);
+    },
+  },
+  computed: {
+    folderParams() {
```
subdavis: think I should maybe roll this and itemParams into `row` since it's the only function that calls them.

Might make the vue computed property jitter less bouncy to have them all in one function too.

src/App.vue
jeffbaumes: https://forum.vuejs.org/t/force-computed-properties-to-recompute/2769

Not sure if this would also solve the problem, but this at least seems a little more simple and explicit (calling `this.$forceUpdate()` in one place seems more readable than having a mysterious `refresh` property that is mentioned in 5 places in the `.vue` file).

subdavis: @jeffbaumes I tried that first.  As far as I can tell, calling $forceUpdate() doesn't have any effect.  I tried calling it with `$refs` from the parent and calling a function in the child that then calls `this.$forceUpdate()` but the computed properties don't re-compute.

EDIT: Poked around and Evan confirms this: https://github.com/vuejs/vue/issues/7395#issuecomment-355721554

Looks like there is actually no way to do this without an explicit dependency update.

jeffbaumes: Here is an officially sanctioned way to deal with this (which matches your approach well enough):

https://github.com/vuejs/vue/issues/214#issuecomment-57822806

However, I believe there is a more fundamental issue going on. It feels like using `asyncComputed` tarnishes the purity of computed properties. Computed properties are automagic and work because they are only supposed to depend on normal data properties. `asyncComputed` properties by definition depend on "other stuff", like server calls which actually hold the state, so we may find ourselves often bumping up against this impurity when using `asyncComputed`. It could be argued that we should not be using `asyncComputed`, but should have normal data properties for things like `count` and `rows`, and methods that mutate them with async code. That way, we would just call the appropriate method(s) when the upload completes, and things will reactively update as normal.

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,294 @@
...
+    changeLocation(item) {
+      const { type, id, name } = item;
+      if ((this.location.id !== id || this.location.type !== type)
+          && type !== 'item') {
+        this.$emit('update:location', { type, id, name });
+      }
+    },
+    requireSession() {
+      if (!this.login || !this.location) {
+        throw new Error('File Browser expects an active session and a defined location.');
```
jbeezley: We should leave open the possibility of browsing anonymously.  As for the `location` prop, we can do a prop validator rather than checking it before every rest call.

src/App.vue
subdavis: I think you may be right about `asyncComputed`.  I do, however, like the organization (particularly the ease of forming a mental model of the dependency tree) it affords you.   @jbeezley do you have any insight about whether or not `asyncComputed` is appropriate in this case, since you've spent a lot more time with it than me?

src/App.vue
```diff
@@ -50,6 +61,33 @@ export default {
...
+          name: this.location.name,
+          _id: this.location.id,
+          _modelType: this.location.type,
+        };
+      }
+      return this.folder;
+    },
+  },
+  methods: {
+    initializeLocation() {
```
jbeezley: Maybe call this `resetLocation` because it's not just called at initialization.

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,294 @@
...
+        return rows;
+      },
+      watch() {
+        return [
+          this.folderParams,
+          this.itemParams,
+          this.counts,
+          this.location,
+          this.refresh,
+        ];
```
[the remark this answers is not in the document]
subdavis: I've found that this de-bounces the async computed properties quite a bit.  With these lines, the app produces many fewer invalid rest requests while the state settles.  It serves to flatten out what was a pretty deep dependency tree.

It also avoids prevents a race condition where the value of counts returns too late so a previous value is used and `rows` ends up containing more rows than pagination allows.

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,294 @@
...
+    rows: {
+      default: [],
+      async get() {
+        this.requireSession();
+        const request = async (gr, endpoint, params) => {
+          if (params === null) return [];
+          const resp = await gr.get(endpoint, { params });
+          if (resp.status !== 200) return [];
+          return resp.data;
+        };
```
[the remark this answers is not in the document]
matthewma7: I agree with Jon. This logic unnecessary and is swallowing exceptions. I think I can just have a plain request and let it fail fast.

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,285 @@
...
+    folderParams() {
+      return {
+        parentType: this.location.type,
+        parentId: this.location.id,
+        limit: this.pagination.rowsPerPage,
+        offset: (this.pagination.page - 1) * this.pagination.rowsPerPage,
+      };
+    },
+    itemParams() {
+      const limit = this.counts.nFolders > this.folderParams.offset
```
matthewma7: I didn't fully understand the logic of the limit calculation, so we could add some comments?

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,285 @@
...
+  computed: {
+    folderParams() {
+      return {
+        parentType: this.location.type,
+        parentId: this.location.id,
+        limit: this.pagination.rowsPerPage,
+        offset: (this.pagination.page - 1) * this.pagination.rowsPerPage,
+      };
+    },
+    itemParams() {
```
matthewma7: Overall, the `itemParams` and `folderParams ` are only being used by the `rows` computed, in addition to the duplicate request behavior I mentioned in another comment, I am thinking maybe we should consider reconstruct the logic and make them simply local variables inside the rows computed, instead of two explicit computed.

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,285 @@
...
+      breadcrumbLoading: false,
+      pagination: {
+        rowsPerPage: 10,
+        page: 1,
+      },
+      selected: [],
+      refreshCounter_: 0, // https://github.com/girder/girder_web_components/issues/39
+    };
+  },
+  asyncComputed: {
```
matthewma7: This is a minor thing, I don't know if it is just me, maybe we should place asyncComputed after computed in general?

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,285 @@
...
+        const request = async (gr, endpoint, params) => {
+          if (params === null) return [];
+          const resp = await gr.get(endpoint, { params });
+          if (resp.status !== 200) return [];
+          return resp.data;
+        };
+        this.rowsLoading = true;
+        const rows = [
+          ...(await request(this.girderRest, GIRDER_FOLDER_ENDPOINT, this.folderParams)),
+          ...(await request(this.girderRest, GIRDER_ITEM_ENDOINT, this.itemParams)),
```
matthewma7: I think this is a place we should use Promise.all([]) instead of two await because following `await`s won't be executed until the former ones finish, losing the async concurrency that we should have.

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,285 @@
...
+  //- Table Row Slot
+  template(slot="items", slot-scope="props")
+    tr.itemRow(:active="props.selected",
+        @click="props.selected = !props.selected",
+        :key="props.index")
+      td
+        v-checkbox.secondary--text.text--darken-1.pr-2(
+            :input-value="props.selected", accent, hide-details)
+      td.pl-1
+        span.text-container.secondary--text.text--darken-3(
```
matthewma7: I didn't find a clean solution excepted for adding line-height to make the text and icon better aligned.

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,285 @@
...
+
+<template lang="pug">
+v-data-table.girder-file-browser-component.elevation-1(
+    select-all,
+    :headers-length="4",
+    v-model="selected",
+    :pagination.sync="pagination",
+    :items="rows",
+    :total-items="totalItems",
+    :loading="loading ? 'accent' : false",
```
matthewma7: This one is totally on Vuetify. But maybe we could overwrite the style so that the loading indicator doesn't push the content down and up when being shown and invisible.

subdavis: I looked into that.  Because of where it's placed in the table, there's just no way to do this.  I'd have to write a new loader that puts the content of the loader into the table header then position it absolutely to keep it from taking up space on the page. 

I'm actually pretty miffed that Vuetify would do something this silly.

matthewma7: It was not straight forward, but seems adding 
```
.girder-file-browser-component .v-datatable__progress .v-progress-linear
  position absolute
```
Would do the trick

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,279 @@
...
+        };
+      },
+      watch() {
+        return [
+          this.pagination,
+          this.location,
+        ];
+      },
+    },
+    rows: {
```
matthewma7: @subdavis Thanks for updating this rows() and the comment on limit calculation. @jbeezley @zachmullen It seems the doubled fetch still exists and its caused by vue-async-computed. The rows() depends on counts(), and it's being evaluated once before counts() updates and once after counts() updates. I created a [jsfiddle ](https://jsfiddle.net/jfmorio/fpyowz6v/1/) to illustrate the probelm. I have been using vue-async-computed in other projects and I like it, but this seems is a case where normal data() + watch:{} is better suited.

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,279 @@
...
+        span.text-container.secondary--text.text--darken-3(
+            :class="{selectable: props.item.type !== 'item'}",
+            @click.stop="changeLocation(props.item)")
+          v-icon(:color="props.selected ? 'accent' : ''") {{ $vuetify.icons[props.item.icon] }}
+          | &nbsp; {{ props.item.name }}
+      td.text-xs-right.secondary--text.text--darken-3(colspan="2") {{ props.item.size }}
+</template>
+
+<style lang="stylus" scoped>
+@import '~vuetify/src/stylus/settings/_colors.styl'
```
matthewma7: Hi @subdavis, I just noticed that you are importing Vuetify stylus in this PR. But the recent master change to SCSS will prevent you from doing this after rebasing master. But at the same time, I think switching to JavaScript color is better because using JavaScript color will make it adaptive to individual app theme and, as Vuetify document states, using stylus color will increase CSS export by~30kb.

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,279 @@
...
+.girder-file-browser-component.elevation-1
+  .selectable
+    opacity .8
+    &:hover
+      opacity 1
+      cursor pointer
+  .v-table tr
+    &.itemRow
+      &[active], &:hover
+        background: $light-blue.lighten-5 !important
```
subdavis: Hey @manthey -- I did this here only because I needed to override color on a tr with `!important` because the theme colors aren't respected by Vueitfy for background active and hover row colors.

As far as I'm aware, there isn't a way to inline pseudo-selectors like `:hover`, so setting the hover and active background to blue can only be done from the stylesheet, where JS colors aren't available.

After rebasing, I think this color import will still be necessary.  Does that sound right, or is there another way?

matthewma7: @subdavis Ah, I see. This is tricky. I was thinking we should always use javascript color to support [theming](https://vuetifyjs.com/en/style/theme) out of box, but seems there are some things I failed to consider. Maybe since Vuetify is moving to SCSS and this usage is static, we just copy the value of $light-blue.lighten-5 for now?

subdavis: Works for me!  Thanks .

matthewma7: Cool, thanks 🙂

jeffbaumes:
@subdavis: @jtomeck would like to try some styling changes, but this is still using stylus. Can you merge in master and update styles to SCSS before he works on styling?

subdavis:
Sure thing.  @jeffbaumes @jtomeck rebased and changed to scss.

subdavis:
Outstanding issues for File Browser:

* Keep using vue-async-computed?
* breadcrumb permissions.  @jeffbaumes mentioned that there may be folders in the path that the user doesn't have access to, and would not be capable of constructing the full breadcrumb.  I think we should just build as much as we can, then only show the parts we were able to build.
* breadcrumb wrap. @jtomeck suggested a while ago that we should only show the most recent ancestors that fit on one line. 

Here's how it works for now, which definitely isn't ideal.

screenshot_20181108_105814


Solved issues:

* The jumpy loading bar - thanks @matthewma7 
* Duplicate requests, solved with data/watch - thanks @matthewma7 again

Otherwise, I think I've addressed the issues mentioned so far.

I'd like to consider bringing in https://alligator.io/vuejs/vue-media-queries/ to collapse the text in the "New Folder" and "New Item" buttons so that their icons are the only thing that show up on mobile, but that would add a dependency.

matthewma7:
Cool, @subdavis Thanks for the changes. 
On performance optimization, the only thing I can think of is maybe creating a simple cache of the breadcrumb requests internally inside the File Explorer, but this is no big deal, we could potentially do this when other parts of this component settled. 

I personally think we should keep vue-async-computed in this project but use it wisely (basically means use it with caution when an async computed depends on another async computed)

About media query, would the utilities provided by Vuetify would work good enough for us? https://vuetifyjs.com/en/layout/breakpoints

subdavis:
> On performance optimization, the only thing I can think of now is maybe creating a simple cache of the breadcrumb requests internally inside the File Explorer

There was a mention of this above, and Jeff replied:

> I personally wouldn't worry about caching and would relax this constraint. Code clarity probably trumps efficiency at this point in the development of these components.

I can do either, but it might be good to let this one slide until it becomes a problem.  IMO, it would be nice to have this as a girder feature rather than fix it on the client.  Perhaps extend the `folder/id/details` endpoint such that it reports the full path.  That would also solve the permissions problem.

subdavis:
@zachmullen @matthewma7 

What other changes would you like to see here in order to get this PR merged?

After this, I have markdown and folder creation, which should be ready for review soon.

matthewma7:
@subdavis All looks good to me. 👍

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,280 @@
+<script>
+import { sizeFormatter } from '../utils/mixins';
+import GirderBreadcrumb from './Breadcrumb.vue';
+
+const GIRDER_FOLDER_ENDPOINT = 'folder';
+const GIRDER_ITEM_ENDOINT = 'item';
```
zachmullen: `ENDPOINT`*

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,280 @@
...
+const ICON_MAP = {
+  folder: 'folder',
+  item: 'file',
+};
```
zachmullen: This is meant to be captured by the icon mapping set up in [vuetifyConfig.js](https://github.com/girder/girder_web_components/blob/3f5ae62664e9f6fa02dbc451e6ccca3c19c1f81b/src/utils/vuetifyConfig.js#L8-L24), so this map feels redundant with that one.

src/components/FileBrowser.vue
```diff
@@ -0,0 +1,280 @@
...
+  },
+  asyncComputed: {
+    counts: {
+      default: { nFolders: 0, nItems: 0 },
+      async get() {
+        const endpoint = `${this.location.type}/${this.location.id}/details`;
+        const { data } = await this.girderRest.get(endpoint);
+        return {
+          nFolders: data.nFolders ? data.nFolders : 0,
+          nItems: data.nItems ? data.nItems : 0,
```
zachmullen: We can use a shorthand that is equivalent to this ternary:

```javascript
{
    nFolders: data.nFolders || 0,
    nItems: data.nItems || 0
}
```

src/components/Breadcrumb.vue
```diff
@@ -0,0 +1,107 @@
...
+v-breadcrumbs.girder-breadcrumb-component.pl-0.pt-0.pb-0
+  v-icon.mdi-24px(
+      :class="{ disabled: disabled }",
+      slot="divider",
+      color="accent") {{ $vuetify.icons.chevron }}
+  v-breadcrumbs-item(
+      :class="{ disabled: disabled }",
+      @click.native="$emit('changelocation', breadcrumb.root)")
+    v-icon.mdi-24px(color="accent") {{ $vuetify.icons.globe }}
+    | &nbsp; {{ breadcrumb.root.name }}
```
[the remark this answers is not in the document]
subdavis: I fought with this for a bit and gave up.  I found that I needed more than just a single-width space here.

0 or more whitespace characters:

screenshot_20181113_161203

nbsp:


screenshot_20181113_161227

However, it never occurred to me to just put some padding on the icon.  `.pr-2` solves my issue here.

src/components/Breadcrumb.vue
```diff
@@ -0,0 +1,107 @@
...
+
+<template lang="pug">
+v-breadcrumbs.girder-breadcrumb-component.pl-0.pt-0.pb-0
+  v-icon.mdi-24px(
+      :class="{ disabled: disabled }",
+      slot="divider",
+      color="accent") {{ $vuetify.icons.chevron }}
+  v-breadcrumbs-item(
+      :class="{ disabled: disabled }",
+      @click.native="$emit('changelocation', breadcrumb.root)")
```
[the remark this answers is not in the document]
subdavis: `crumbclick` has a nice ring to it.

src/components/Breadcrumb.vue
```diff
@@ -0,0 +1,107 @@
...
+    v-icon.mdi-24px(color="accent") {{ $vuetify.icons.globe }}
+    | &nbsp; {{ breadcrumb.root.name }}
+  v-breadcrumbs-item(
+      :class="{ disabled: disabled }",
+      v-for="item in breadcrumb.path",
+      :key="`${item.id}.crumb`",
+      @click.native="$emit('changelocation', item)") {{ item.name }}
```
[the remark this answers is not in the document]
subdavis: `@click` doesn't work on any Vuetify components because they aren't native DOM elements.

I think elements like v-btn step around this by forwarding native events to the parent -- see https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VBtn/VBtn.ts#L104

For whatever reason, the event isn't forwarded for v-breadcrumb-item, so we can bind to the root element's events directly with `.native`.

https://github.com/vuetifyjs/vuetify/issues/4730

src/components/Breadcrumb.vue
```diff
@@ -0,0 +1,107 @@
...
+  &.v-breadcrumbs li:nth-child(2n) {
+    padding: 0 4px !important;
+  }
+
+  li.disabled a,
+  li.v-breadcrumbs__divider i.disabled {
+    color: gray !important;
+    cursor: default !important;
+  }
+}
```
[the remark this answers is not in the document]
subdavis: For this I'm going to tag in @jtomeck 

Those styles don't affect the FileBrowser, they're for the breadcrumb in the Create Folder popup where the breadcrumb is static.  That PR is #48 

In general, it seems useful to be able to "disable" the breadcrumbs so they don't appear interactive.

```
v-breadcrumbs.headline
```

This overrides the typography set in App.vue, and would prevent consumers of this library from setting global font themes.  It doesn't look like Vuetify supports any font other than Roboto.

The font themes @jtomeck wants to use is "Barlow Condensed"

src/components/FileBrowser.vue
subdavis: This currontly maps from girder `_modelType` to an icon described in vuetifyConfig.

My motivation here was to allow us to switch the icon mappings without having to refactor other parts of the code that may happen to use the same icons.

I believe it may be useful to separate our global icon map from the "filetype" map, which will probably get quite long as we add icons for images, xml, json, binary, and whatever else.  In the future, I think this map will also include mime types or extensions.

That being said, this may also be YAGNI.

src/components/Breadcrumb.vue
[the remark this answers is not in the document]
jeffbaumes: > This overrides the typography set in App.vue, and would prevent consumers of this library from setting global font themes. It doesn't look like Vuetify supports any font other than Roboto.

The proper way to change Vuetify's font, I believe, is to update stylus variables in a custom `main.styl` described [here](https://vuetifyjs.com/en/style/theme#stylus-guide).

There is the additional problem that headings (like headline) are dealt with separately. I have been able to make this [main.styl](https://gist.github.com/jeffbaumes/5ea3ae818a8cae4bcb64579ef2487d44) change the font for all things, including headings. Redefining `$headings` is not ideal (copy/paste from Vuetify stylus) but I could not seem to get it work without it.

It also requires stylus webpack rules, but it seems Vuetify is [moving to SCSS](https://github.com/vuetifyjs/vuetify/issues/5558) in the future, so it would be a temporary thing to support both in this example app bundle.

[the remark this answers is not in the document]
zachmullen: > This overrides the typography set in App.vue

I don't understand what you mean, isn't the `headline` class part of Vuetify's typography API?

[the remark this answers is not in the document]
jeffbaumes: What happens is that just setting the font on the body does not override all of Vuetify's fonts. There is a more specific CSS for `headline` in Vuetify that is Roboto and "overrides" the customized Barlow Condensed body CSS font in App.vue (which in turn was overriding Vuetify's body font of Roboto). My comment addresses this by explaining a way to truly change Vuetify's default font, so that Roboto no longer appears in the CSS, both for body text and all headings.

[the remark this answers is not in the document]
zachmullen: Ah, thanks for the explanation. I assumed "headline" and the related typography classes mostly controlled size, weight, color, rather than typeface.

@jtomeck how important is it for us to override Vuetify's prescribed typeface? If we want to do it, I'd like to make sure we do it the way @jeffbaumes describes as well so that we can use Vuetify's typography classes rather than having to use manual, ad hoc CSS rules for those things, in the interest of using a consistent theme that is easy to control centrally.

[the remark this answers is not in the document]
subdavis: I'm personally not a huge fan of "Barlow Condensed".  I find myself squinting at it.  

We could certainly scale it up, but that would (i think) involve changing roughly 25 font sizes from [variables](https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/stylus/settings/_variables.styl) that have already been optimized for Roboto.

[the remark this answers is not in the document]
jeffbaumes: I'll point out that the font can be changed entirely through Vuetify Stylus/CSS customization. Our library of components should not change the font internally; apps should be able to decide what to do with Vuetify. There are several choices to make: 1. What font do we want our example app showcase to have? 2. What font do we want the new Girder UI to have? 3+. For project X, what font does it want to have? The answers to these should be able to be different for each use case, though the answers to 1 and 2 will probably set precedent for downstream projects.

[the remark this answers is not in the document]
matthewma7: My two cents here. The Barlow Condensed feels a little bit too condensed for me if used everywhere, meanwhile default Roboto do use a lot of space in actual use. So, maybe we could try the Roboto Condensed as suggested by @jtomeck? So that when Project X decides to use default Roboto, it won't be too big a change.

src/components/Breadcrumb.vue
```diff
@@ -0,0 +1,106 @@
...
+      :key="item.id") {{ item }}
+</template>
+
+<style lang="scss">
+.girder-breadcrumb-component {
+  &.v-breadcrumbs li:nth-child(2n) {
+    padding: 0 4px !important;
+  }
+
+  li.disabled a,
```
zachmullen: `v-breadcrumbs-item` has a `disabled` attribute, is there a reason not to just use that instead?

src/components/FileBrowser.vue
zachmullen: Is the intent that an *item* in this widget will be shown with different icons based on data type of the item? We don't formally encode the data type of an item like we do with files, so we'd need to come up with a system of rules to map an item to an icon based on some metadata in the item.

I think until we come up with that, it's probably premature to have a second level of indirection since we are only dealing with folder and item, which should be in the general icon map.

src/components/Breadcrumb.vue
```diff
@@ -0,0 +1,106 @@
...
+      color="accent") {{ $vuetify.icons.chevron }}
+  v-breadcrumbs-item(
+      :class="{ disabled }",
+      @click.native="$emit('crumbclick', breadcrumb.root)")
+    v-icon.mdi-24px.pr-2(color="accent") {{ $vuetify.icons.globe }}
+    | {{ breadcrumb.root.name }}
+  v-breadcrumbs-item(
+      :class="{ disabled }",
+      v-for="item in breadcrumb.path",
+      :key="`${item.id}.crumb`",
```
zachmullen: Can change this key to just `item.id`

src/components/Breadcrumb.vue
```diff
@@ -0,0 +1,106 @@
...
+    location: {
+      type: Object,
+      required: true,
+      validator: val => val.type && val.id,
+    },
+    disabled: {
+      type: Boolean,
+      default: false,
+    },
+    append: {
```
zachmullen: Doesn't look like this is used, what's the reason for this prop?

subdavis: It's used in #48

subdavis:
Are there any other changes required for this to merge?

Since fonts will affect every component, could we create a different issue for library-wide font configuration and dealing with that in another PR?  

I don't think the discussion around fonts is tightly coupled with this particular PR.

src/utils/main.styl
```diff
@@ -0,0 +1,3 @@
+@import url('https://fonts.googleapis.com/css?family=Barlow+Condensed:300,300i,400,400i,600,600i');
```
jeffbaumes: Do we need this line (or even this file) anymore since we're punting on fonts for this PR?

jeffbaumes:
I agree @subdavis. I don't think we should do fonts on this PR but should have an issue where we address this later on. I made one comment on removing a `styl` file which might be the remnants of this.

src/utils/main.styl
subdavis: We do not!

zachmullen:
> Are there any other changes required for this to merge?

Did we resolve the issue of debouncing / multiple requests happening unnecessarily when things are updated?

src/components/index.js
```diff
@@ -1,9 +1,14 @@
...
+
 import Authentication from './Authentication/';
+import Breadcrumb from './Breadcrumb.vue';
+import FileBrowser from './FileBrowser.vue';
 import Upload from './Upload.vue';
 
 export {
-  Upload,
   Authentication,
+  Breadcrumb,
```
zachmullen: My vote would be that we do not expose this component in index.js, at least not at this time. It seems like it's more of an internal helper component for some of our other components, rather than something that will stand alone.

I could be convinced otherwise, if people feel differently. @jeffbaumes @jbeezley @subdavis

zachmullen:
Sorry to wait until the very last moment to say this, but could we rename this component to "DataBrowser" or something a little more generic, since "File" refers to a specific concept in Girder that doesn't appear in this component?

src/components/index.js
jbeezley: Yes, I would agree minimizing the api surface until we have a need to expose it.

tests/unit/utils.js
```diff
@@ -1,10 +1,9 @@
-import { createLocalVue } from '@vue/test-utils';
+import Vue from 'vue';
 import Girder from '@/index';
 
 const girderVue = () => {
-  const localVue = createLocalVue();
-  localVue.use(Girder);
-  return localVue;
+  Vue.use(Girder);
```
zachmullen: I'm fine with this change, but was it necessary?

package.json
```diff
@@ -49,7 +49,8 @@
     ],
     "rules": {
       "no-underscore-dangle": 0,
-      "import/prefer-default-export": 0
+      "import/prefer-default-export": 0,
+      "no-await-in-loop": 0
```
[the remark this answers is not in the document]
zachmullen: Sorry, by "leave this rule in place" I meant remove the disabling of this rule, I realized that was totally unclear :)

src/components/Breadcrumb.vue
```diff
@@ -0,0 +1,100 @@
...
+      @click.native="$emit('crumbclick', item)") {{ item.name }}
+  v-breadcrumbs-item(
+      v-for="item in append",
+      :key="item.id") {{ item }}
+</template>
+
+<style lang="scss">
+.girder-breadcrumb-component {
+  &.v-breadcrumbs li:nth-child(2n) {
+    padding: 0 4px !important;
```
zachmullen: I'm still in favor of removing this rule, I don't think the padding is important enough for us to be reaching deep into Vuetify internals to change it. I think our best bet would be to open an issue request (or even a PR!) against Vuetify to allow a `dense` prop on the breadcrumb component, or make it so that the `divider` slot is in charge of its own padding (which may be harder to pitch to them since it's breaking).

src/components/Breadcrumb.vue
```diff
@@ -0,0 +1,100 @@
...
+  },
+};
+</script>
+
+<template lang="pug">
+v-breadcrumbs.girder-breadcrumb-component.font-weight-bold.pa-0
+  v-icon.mdi-24px(
+      :disabled="disabled",
+      slot="divider",
+      color="accent") {{ $vuetify.icons.chevron }}
```
zachmullen: The "accent" color makes these dividers look like links, which they aren't. The default is that they are grey, which I think I like better. @jtomeck are you OK to change the color?

zachmullen:
Right now the "select" (click the row without clicking the name) causes the checkbox to be checked. I think selecting and checking a row should be independent actions.

subdavis:
>  think selecting and checking a row should be independent actions

Could you link me to an example of where a data table behaves like this in the wild?  I'm having trouble understanding the difference between select and check.

Do you mean that there should be 2 events published by the DataBrowser?  Independent lists for "selected" and "checked"?  What is the difference in purpose between select and check (i.e. when would I want to do one vs. the other?)

zachmullen:
GMaail is one example, clicking the row actually changes the route, not check the box. But in our case, there are really 3 different actions people want to take on a row:

1. Drill down into that resource (i.e. descend the hierarchy)
2. Check some items (for batch operations)
3. Select a resource or multiple resource (e.g. "select a folder", in which case we wouldn't render the checkbox at all).

I could be wrong about this behavior, curious was @jtomeck thinks

src/components/Breadcrumb.vue
subdavis: I'd prefer not to have this as well, but I find the spacing between dividers (12px on either side) to be unreasonably large.  The actual problem isn't the padding -- it's the mdi-chevron icon which makes really poor use of its vector box - it has another 12px between the wall of the image box and the start of the icon, making the gap look like ~24px to either side.

I don't think this is a large enough use case for the Vuetify maintainers to accept a `dense` prop as compensation for an abnormal icon, and fixing the icon doesn't seem like a great option either.

zachmullen: What about if we just use a `/` or a [unicode chevron](https://www.compart.com/en/unicode/U+203A)?

zachmullen: I like the slash too, and it also will be familiarly associated with the filesystem separator.

subdavis:
Updated to use `/` as separator.  I'm aware that this is the default separator, and leaving this slot empty would look almost the same.  But the `subheading` style makes `/` noticeably bigger, which makes the bottom of the character extend below the baseline and looks a lot better to me.

Here it is without the slot:

screenshot_20181116_154856

I think we should keep the custom styled `/`.

zachmullen:
That's fair. It's worth mentioning that the `color="accent"` on the divider element no longer has any effect, so we can remove it.

subdavis:
I may be getting too personally invested in the minutia of DataBrowser.  We've been together for so long now.  I nearly fell on my sword over the height of a `/`.

We can totally remove the custom slot.

jeffbaumes:
@subdavis we as a team/company tend to underestimate the importance of good clean design, so please don't feel you are being overly picky. I'm also not saying we should keep the custom slot because there is always a maintainability tradeoff, but just know that attention to design detail is generally something we should strive to improve.

src/components/Breadcrumb.vue
jtomeck: @zachmullen I don't know that it's super important for me to push back on because it really just comes down to opinion, so if you really think it's needed I'm fine with changing the color. I'll give my opinion on it anyway though in case anyone's interested :P

From a designer's perspective it seems like many people are of the mindset that giving something a color makes it look like a link. I think it all depends on the usage. For example, if I had a block of copy and changed the color of a few words, sure they would look like a link. I don't think things like giving a prominent headline a color, or giving dividers a color make them look like links, especially since nothing happens on hover and the cursor does not change.

jtomeck:
> Right now the "select" (click the row without clicking the name) causes the checkbox to be checked. I think selecting and checking a row should be independent actions.

I agree with this ^

> * Drill down into that resource (i.e. descend the hierarchy)
> * Check some items (for batch operations)
> * Select a resource or multiple resource (e.g. "select a folder", in which case we wouldn't render the checkbox at all).

As well as this ^

jtomeck:
> I may be getting too personally invested in the minutia of DataBrowser. We've been together for so long now. I nearly fell on my sword over the height of a `/`.

Now you know the plight of the designer @subdavis haha

subdavis:
> removal of no-await-in-loop

Done.  Thought I did that already...

> `&nbsp; -> .pl-1 or something`

:+1: 

> Right now the "select" (click the row without clicking the name) causes the checkbox to be checked. I think selecting and checking a row should be independent actions.

I gave this a try.  Requiring the checkbox click for `check` makes sense, but having any other part of the row trigger `select` felt odd.  I could click any part of the row and navigate into a subfolder, meaning a stray click anywhere on the table has the potential to disorient a user: "Hey, what changed? What did I do?"

Should this table behave like the current Girder browser, where a stray click on some part of the table that isn't an actionable item simply do nothing?  

BTW: I modeled my original behavior on [this table from the Vuetify docs](https://codepen.io/pen/?editors=1010).

zachmullen:
Barring any objections, I think we should merge this. If there are any lingering issues, let's open tickets for them. A couple things that might merit issues:

* Multiple requests getting sent unnecessarily... was this resolved on this branch or is it still an issue?
* Check vs. select actions

subdavis:
> Multiple requests getting sent unnecessarily... was this resolved on this branch or is it still an issue?

This was resolved thanks to guidance from @matthewma7 

> Check vs. select actions

Opened #51

zachmullen:
:shipit:

PR #37, opened 2018-10-05
Introduce vue-recommended eslint rules

zachmullen:


zachmullen:
This still doesn't capture all of the rules in the styleguide, but it should now track them modulo the upstream release cycle.

zachmullen:
Rebased, this is ready for review.

Issue #41, opened 2018-10-15
Use SCSS instead of Stylus


matthewma7:
Hi, I know Girder has been using Stylus. However, Vuetify is [moving away from Stylus to SCSS in 2.0](https://www.patreon.com/posts/september-2018-21931872), which I guess is motivated by a [feature request](https://github.com/vuetifyjs/vuetify/issues/2371).
And as a core library, I think if we could ask people to learn as little as possible to be able to understand the code, create PR, or hack the code it is better. SCSS is a superset of CSS, so anyone knows CSS would start to write or modify style with the minimal learning curve.

zachmullen:
> And as a core library, I think if we could ask people to learn as little as possible to be able to understand the code, create PR, or hack the code it is better. SCSS is a superset of CSS, so anyone knows CSS would start to write or modify style with the minimal learning curve.

Stylus is also a superset of CSS, so we'll need a different argument to motivate a switch. That thread you linked seems to indicate their interest in moving has to do with stylus not being as well supported, which might be compelling. Is there anything other than that that would make the switch attractive to our community that already has a lot of inertia with Stylus? I am not familiar with SCSS, maybe there are some cool features not present in stylus?

matthewma7:
Feature-wise, it sounds like SASS/SCSS may be a little better, but I don't know without looking into it.
So, other than feature and support the [general fondness and prevalence of SCSS](https://2017.stateofjs.com/2017/css/results) could be a reason.
Except for inertia, does Girder community like Stylus for some particular reason? I was thinking girder 3 and this component library is a good chance for changing.

zachmullen:
Popularity is definitely a reason to consider a switch. As long as there isn't a big cost to switching from stylus to scss, it might be worthwhile just for that reason.

matthewma7:
I agree. And the cost is small at this moment because the code base here is small.

zachmullen:
A PR that shows the code changes required to make the switch (even if just scoped to a single existing component) is probably the best way to push this forward. Feel free to make such a PR if you think it's an easy change.

matthewma7:
It should easy, I will create such PR.

jeffbaumes:
This does indeed look a bit grim:

https://github.com/stylus/stylus/graphs/contributors

This (ad-ridden) page seems to do a decent job summing up pros and cons of the two:

https://www.slant.co/versus/762/764/~stylus_vs_sass

* Stylus: clean syntax but not maintained.
[...]

PR #42, opened 2018-10-18
Switch to SCSS

matthewma7:
This resolve issue https://github.com/girder/girder_web_components/issues/41

I added linting ability as suggested by @zachmullen, but I failed to make the configuration part of the package.json like other options do. (Probably needs to modify the plugin  to support that)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "build": "vue-cli-service build --target lib --name girder src/index.js",
     "lint": "vue-cli-service lint",
     "lint:pug": "pug-lint-vue src",
+    "lint:style": "stylelint 'src/**/*.vue'",
     "test:unit": "vue-cli-service test:unit"
   },
   "dependencies": {
@@ -32,6 +33,8 @@
     "node-sass": "^4.9.0",
     "pug-lint-vue": "^0.1.3",
     "sass-loader": "^7.0.1",
+    "stylelint": "^9.6.0",
+    "stylelint-config-standard": "^18.2.0",
     "vue-cli-plugin-pug": "^1.0.7",
     "vue-template-compiler": "^2.5.16"
   },
@@ -116,6 +119,11 @@
     "validateLineBreaks": "LF",
     "validateSelfClosingTags": true,
     "validateTemplateString": true
+  },
+  "stylelint": {
+    "extends": [
+      "stylelint-config-standard"
+    ]
   },
   "browserslist": [
     "> 1%",
--- a/package.json
+++ b/package.json
@@ -29,9 +29,9 @@
     "axios-mock-adapter": "^1.15.0",
     "babel-core": "7.0.0-bridge.0",
     "babel-jest": "^23.6.0",
+    "node-sass": "^4.9.0",
     "pug-lint-vue": "^0.1.3",
-    "stylus": "^0.54.5",
-    "stylus-loader": "^3.0.2",
+    "sass-loader": "^7.0.1",
     "vue-cli-plugin-pug": "^1.0.7",
     "vue-template-compiler": "^2.5.16"
   },
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,6 +17,7 @@ jobs:
             - ~/.cache/yarn
       - run: yarn lint --no-fix
       - run: yarn lint:pug
+      - run: yarn lint:style
       - run: yarn test:unit --coverage --coverageReporters html
       - store_artifacts:
           path: coverage/
--- a/src/App.vue
+++ b/src/App.vue
@@ -53,9 +53,10 @@ export default {
 };
 </script>
 
-<style lang="stylus">
-.app
-  font-family 'Avenir', Helvetica, Arial, sans-serif
-  -webkit-font-smoothing antialiased
-  -moz-osx-font-smoothing grayscale
+<style lang="scss">
+.app {
+  font-family: "Avenir", Helvetica, Arial, sans-serif;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
 </style>
--- a/src/components/Upload.vue
+++ b/src/components/Upload.vue
@@ -167,67 +167,90 @@ export default {
 };
 </script>
 
-<style lang="stylus" scoped>
-$stripeColor = #f0f0f3
-$img = linear-gradient(
-  -45deg, $stripeColor 25%, transparent 25%, transparent 50%, $stripeColor 50%,
-  $stripeColor 75%, transparent 75%, transparent)
+<style lang="scss" scoped>
+$stripeColor: #f0f0f3;
+$img: linear-gradient(
+  -45deg,
+  $stripeColor 25%,
+  transparent 25%,
+  transparent 50%,
+  $stripeColor 50%,
+  $stripeColor 75%,
+  transparent 75%,
+  transparent
+);
 
-.dropzone-wrapper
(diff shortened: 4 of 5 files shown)


zachmullen:
Doesn't scss have an alternative syntax that lets us omit the curly braces? I'd like to use that if so, unless there's some reason the prefer this syntax.

matthewma7:
The reasons I can think of are, one familiarity: I am more used to see scope with braces instead of with indentation. I feel this way it is natural when working alongside with javascript. and two interoperability: the SCSS syntax is a superset of CSS on syntax, so it is very easy for me to copy to/from browser dev tools in my daily development, and copy from some example code online sometimes. Overall, for my work habit, omitting colons and/or semicolons always meant more work for me.

zachmullen:
Very good points. Combined with the fact that we are attempting to minimize the amount of CSS we write anyway, I think the syntax is less of a concern than other important factors, so I'm on board with this change.

matthewma7:
Hi @zachmullen, If the PR looks good, should we approve it?

jbeezley:
I don't feel strongly either way.  I'm in favor of switching to SCSS because it is more popular.  I mildly prefer the stylus-like syntax, but not enough to object.

.stylelintignore
```diff
@@ -0,0 +1 @@
+# .gitignore syntax (uses node-ignore behind the scenes)
```
zachmullen: Why is this file necessary?

package.json
```diff
@@ -17,9 +18,10 @@
     "qs": "^6.5.2",
     "vue": "^2.5.16",
     "vue-async-computed": "^3.4.1",
-    "vuetify": "^1.1.9"
+    "vuetify": "^1.3.0"
```
zachmullen: This change doesn't have any effect, does it?

package.json
```diff
@@ -29,11 +31,13 @@
...
+    "node-sass": "^4.9.0",
     "pug-lint-vue": "^0.1.3",
-    "stylus": "^0.54.5",
-    "stylus-loader": "^3.0.2",
+    "sass-loader": "^7.0.1",
+    "stylelint-config-standard": "^18.2.0",
     "vue-cli-plugin-pug": "^1.0.7",
-    "vue-template-compiler": "^2.5.16"
+    "vue-template-compiler": "^2.5.16",
+    "vuetify-scss": "^1.2.0"
```
zachmullen: Is this library necessary? We import the vuetify CSS already built, and we don't want to have to know about the upstream build toolchain unless we really have to.

vue.config.js
```diff
@@ -1,3 +1,8 @@
 module.exports = {
   lintOnSave: false,
+
+  pluginOptions: {
+    lintStyleOnBuild: false,
+    stylelint: {},
```
zachmullen: Aren't both of these the defaults? I'd vote to remove them if so.

stylelint.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  root: true,
+  extends: [
+    'stylelint-config-standard',
+  ],
+};
```
zachmullen: It sounds like this could be added as a "stylelint" property under package.json, if so we should do that.

matthewma7: Thanks, the key stylelint works.

vue.config.js
matthewma7: Yes, we can remove these. (But seems the pluginOptions needs to stay, or the @ascendancyy/vue-cli-plugin-stylelint will throw exception)

vue.config.js
```diff
@@ -1,3 +1,6 @@
 module.exports = {
   lintOnSave: false,
+
+  pluginOptions: {
+  }
```
zachmullen: We can just omit this block since it's now empty

package.json
matthewma7: I think I accidentally upgraded Vuetify somehow, but I think we probably should update Vuetify in a probably PR for its fixes and new features.

vue.config.js
matthewma7: It looks like if I omit the key. The lint:style will throw an error `TypeError: Cannot destructure property `lintStyleOnBuild` of 'undefined' or 'null'.
    at module.exports (/home/matthew/girder_web_components/node_modules/@ascendancyy/vue-cli-plugin-stylelint/index.js:7:11)`

package.json
zachmullen: Agreed, since the upgrade of Vuetify is out of scope of this PR, can you also back out those changes from the yarn.lock? We can add it on a second PR and then reviewers will know to poke around the components to make sure everything looks nice.

matthewma7: Great catch. I rebased again to try to clean up the yarn.lock.

package.json
```diff
@@ -117,6 +120,12 @@
     "validateSelfClosingTags": true,
     "validateTemplateString": true
   },
+  "stylelint": {
+    "root": true,
```
zachmullen: The [documentation](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md) doesn't mention `root`, what does this do?

zachmullen:
Just to try and distill down the arguments on this subject to make it easy for people to decide on this issue:

**Pros**
* Much more popular tool, more community support
* Support for stylesheet linting for our Vue components (currently impossible with Stylus)
* Plain-old-CSS syntax makes it easier to copy paste, and is more familiar for new contributors

**Cons**
* Requires a much larger, native build toolchain to be installed by npm. It's only a dev dependency, though, and it installs cleanly on both Debian and MacOS in my testing. The upside of that is that it ought to build our stylesheets faster, though that will likely be a negligible difference.
* No more shorthand syntax. To be fair, there were some uncommon cases in which the stylus syntax could cause an ambiguous parse, so at least we won't run into that.

package.json
matthewma7: I personally didn't know. It was generated for me. The eslintConfig configuration section also has the root. Maybe it has something to do with vue-cli? I didn't find any document as well. https://github.com/vuejs/vue-cli/blob/1a8394434ab074d6d15412f17d8374e84cee9314/packages/%40vue/cli/lib/Generator.js#L99

zachmullen: I just read through the source of `@ascendancyy/vue-cli-plugin-stylelint`, and it seems pretty rough, and IMO unnecessary. It's a whole lot of code that does almost nothing beyond this:

`stylelint src/**/*.vue`

So my preference would be to drop `@ascendancyy/vue-cli-plugin-stylelint` and do

`"lint:style": "stylelint 'src/**/*.vue'"`

If we anticipate including CSS in files aside from just vue single file components, we can add more extensions to that glob expression.

zachmullen: If we cared about linting on build, I'd say it's worth using the plugin, but I'm not a fan of linters causing build errors.

matthewma7: Sounds good, I'll give it a try and see if everything works well without it.

matthewma7: Everything works great, include the autofix that I would use. I'll push the change.

matthewma7: Hi @zachmullen, this PR is ready for further review

zachmullen:
I approve of these changes. I'd like to get @jbeezley @subdavis and @jeffbaumes to also approve it before we merge.

jeffbaumes:
I approve.

jbeezley:
LGTM

matthewma7:
Thanks, Zach and everyone 👌

zachmullen:
Thank you @matthewma7 for the change!

PR #57, opened 2018-12-03
Add hooks

subdavis:
This PR introduces a pattern for single-purpose components like upload and create-folder which add synchronous pre- and post- hooks to the main function of each. 

Some use cases for this:
* Making sure the consumer component has time to properly render animations like a closing dialog animation before the child resets itself.
* Allowing the consumer to update any other children with the results of the dialog action, e.g. update the file browser while the dialog is closing.
* Allow the consumer to inject arbitrary actions into the child dialog flow.  Example: in the UTM demo, I would _really_ like to lazy-create the upload destination folder right before data is uploaded.  I don't want to make the user create the directory by themselves, and I also don't want to spam their filespace with empty folders that get created every time they open the demo page. 

Other things in this pull request:

* add `set` operation for girderRest.token, since, as with demo_site, login could be managed by something other than GWC.
* add toggle for checkboxes in DataBrowser.   Some applications of DataBrowser will have no use for select functionality, so in those cases, checkboxes would just confuse the user and can be turned off.'

Why are these seemingly unrelated things in the same PR?

I've been working on this and demo_site in parallel and experimenting with changes to GWC as they make sense for demo_site.  I can separate them if this is an issue.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -11,13 +11,13 @@ v-app.app
       girder-upload(
           v-if="uploadDest",
           :dest="uploadDest",
-          :hooks="uploadHooks",
+          :post-upload="postUpload",
           multiple="multiple")
     v-dialog(v-model="newFolder", full-width, max-width="800px")
       girder-upsert-folder(
           :key="location._id",
           :location="location",
-          :hooks="upsertHooks",
+          :post-upsert="postUpsert",
           @dismiss="newFolder = false")
     v-card
       girder-data-browser(ref="girderBrowser",
@@ -51,8 +51,6 @@ export default {
       newFolder: false,
       browserLocation: null,
       forgotPasswordUrl: '/#?dialog=resetpassword',
-      uploadHooks: { postUpload: this.postUpload },
-      upsertHooks: { postUpsert: this.postUpsert },
     };
   },
   asyncComputed: {


matthewma7:
_This may be too defensive. I don't think errors in hook functions should stop execution in the component, but perhaps they should?_

This is a great question. As a library, it makes sense to catch it and continue its execution, but I personally lean towards letting things fail fast. 
And, It may not be needed to warp in promise because everything is awaitable. 
I think, if needed, some mixin as helper function internally make sense, but could keep the hook purely internal.

subdavis:
> Everything is awaitable

Huh.  It works in a browser but not in node 10.9

```javascript
a = () => {}
await a()
```

Browser: `undefined`
Node: `SyntaxError: await is only valid in async function`.

None of that actually matters because babel transpiles all the async/await stuff to ES5 with regenerator-runtime, but I guess it's still good to think about.  TLDR: you're right no need to wrap.

> I think, if needed, some mixin as helper function internally make sense, but could keep the hook purely internal.

Sure, lets make the hooks regular props and call them with a hook mixin.

matthewma7:
Hi @subdavis, the latest changes look good, but looks like the App.vue is not up to date in this PR.

subdavis:
@matthewma7 right you are.  Doing a couple things at once and got sidetracked yesterday.  Thanks.

src/App.vue
```diff
@@ -8,16 +8,17 @@ v-app.app
           :oauth="true",
           :forgot-password-url="forgotPasswordUrl")
     v-dialog(v-model="uploader", full-width, max-width="800px")
-      girder-upload(v-if="uploader",
+      girder-upload(
+          v-if="uploadDest",
```
subdavis: @matthewma7 uploadDest is a computed property derived from an asyncComputed property, and so on page load it may be null.  This was not a problem before because by the time uploader toggled on, this computed prop had loaded.  Now the uploader is created on load and must wait.

I'm not sure if we should have the GWC components tolerate invalid props for this reason.  A DataBrowser or Uploader without a valid location is useless, and right now it fills the console with errors.  I think these errors are useful.  

On the other hand, providing the results of asyncComputed properties to GWC components is a very common use case for me so far, so `v-if="locationVariable"` seems to be necessary everywhere I've used GWC.  Are we ok with that?

src/components/DataBrowser.vue
```diff
@@ -207,12 +207,12 @@ v-data-table.girder-file-browser-component.elevation-1(
   //- Table Row Slot
   template(slot="items", slot-scope="props")
     tr.itemRow(:active="props.selected",
-        @click="props.selected = !props.selected",
+        @click="if (selectEnabled) props.selected = !props.selected;",
```
matthewma7: Sometimes, it's better not even bind some event because some Vuetify component changes behavior when there is a handler. But it should be fine in this case, and I can't think of a cleaner way of doing so.

src/App.vue
matthewma7: As long as it doesn't break the animation, it is totally ok. What is GWC?

src/rest.js
```diff
@@ -39,6 +39,10 @@ export default class RestClient extends Vue {
     }));
   }
 
+  set token(val) {
+    this.token = val;
```
zachmullen: I'm having a hard time wrapping my mind around this, what's going on here? At face value it would seem to be an infinite recursion.

subdavis: Frankly I'm not sure what happened here.  Can we change the private version to `_token`?

zachmullen: We could, but what is the purpose of having the setter?

subdavis: https://github.com/girder/demo_site/blob/utm/src/containers/UtmContainer.vue#L39

subdavis: > won't that work without a setter

I just intended to be explicit about allowing external sources to set that property.  In the moment, that seemed like a really important thing to do.  You're right, this can be removed.

PR #59, opened 2019-01-16
Search Widget

### LINKED ISSUE #40: Search Component
I have put together some screens for a search component. The consensus from people that I've talked to is that some/much of this can not be done. I wanted to post anyway to share the ideas and discuss the parts that can be done, as well as potential ways to improve upon how search is currently done in Girder.

01 file list
Fixed position search bar on the bottom of the screen

02 auto complete
Autocomplete

03 advanced search
Clicking the icon next to search (advanced search) opens a dialogue/bottom sheet style component. The fields are just placeholders that I stole from Google Drive. My thought is that there could be tabs to cover different search modes. More modes could be added, ie. for QIDW, we needed a search specifically for DICOM metadata.

04 results
Results page similar to hierarchy widget but stripped down. Paths to files are shown and this list could be broken down based on what type of result is displayed (ie - items/folders/users/etc.)

Let's discuss

subdavis:
screenshot_20190116_125348
screenshot_20190116_125406

Changes in the PR:
* Add search box with results and options popover menus.
* Refactor data table from DataBrowser into its own component so that it can be used to render any data list from girder, especially search results.

This search bar expands to the size of its container, so it could be as large or small as we want it.

Still TODO:
* Add triangle/caret to options menu so that it points to the settings button.
* Add tests for Search view.

fixes #40


src/main.js
```diff
@@ -8,7 +8,14 @@ const girderRest = new RestClient({ apiRoot: 'http://localhost:8080/api/v1' });
 
 girderRest.fetchUser().then(() => {
   new Vue({
-    render: h => h(App),
+    propsData: {
+      foo: 2,
```
subdavis: Not sure what happened here.  Need to clean this up.

subdavis:
Other outstanding questions:

* What to do with errors that arise during search?  Should these be presented to the user, and how?

subdavis:
@jtomeck Good news and bad news.

screenshot_20190117_111354

Good news: Putting the triangle on the button slot works!  Vuetify even makes running the same animation as the menu popover easy by wrapping the triangle in `v-fade-transition`.

Bad news is that, even though the card is at `z-index: 8` and the arrow is at `z-index: ∞`, the shadow from the card still shows on top of the triangle.  I don't know how to fix this.

jtomeck:
@subdavis - The issue is the order of the DOM elements you're trying to z-index. The only way to get the z-index to work properly is to move the menu so that it shares the same parent element as the button (since the menu resides outside of `.application--wrap` altogether).

I'm guessing this can't be done, so I'm proposing a stylized "hack" that makes it look a bit more intentional.

screen shot 2019-01-17 at 11 40 30 am

Thoughts?

subdavis:
@jtomeck that makes sense to me.  I was never really clear on z-index hierarchy.

Your dark-theme hack is one option.  Another would be to just modify the shadow on the top of those boxes.  It's totally invisible on the blue background anyway.  Both involve overriding vuetify styles in a less-than-ideal way, so I think I'd personally prefer something more subtle.

jtomeck:
Yes, this is my biggest frustration with Vuetify. It looks good out of the box, but if you want to make any customizations it is extremely difficult.

That said, what do you have in mind for something more subtle?

jtomeck:
@subdavis - Sorry I didn't see the screenshots. I prefer the white menu, but it if forces you to fight vuetify too much, we can go dark.

src/components/Search.vue
```diff
@@ -0,0 +1,135 @@
...
+      default: true,
+    },
+  },
+  inject: ['girderRest'],
+  data() {
+    return {
+      errtext: '',
+      searchtext: '',
+      searchmode: 'prefix',
+      searchtypes: ['user', 'folder', 'item'],
```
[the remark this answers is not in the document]
subdavis: Not necessarily.  I can update these with the conventional style once #61 is merged.

src/components/Search.vue
```diff
@@ -0,0 +1,135 @@
+<template lang="pug">
```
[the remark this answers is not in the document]
subdavis: I forgot to move it.  Will update this after #61 is merged.

subdavis:
I think the search component still needs arrow-key navigation (and tab navigation).

I'm going to add that as well before requesting that this be merged.

src/components/Search.vue
```diff
@@ -0,0 +1,149 @@
...
+
+<template lang="pug">
+v-layout.girder-searchbar(row, align-center)
+  v-icon.mdi-24px(color="white") {{ $vuetify.icons.search }}
+  v-menu.grow.mx-3(
+      offset-y,
+      content-class="girder-searchbar",
+      :value="searchText",
+      :nudge-bottom="6")
+    v-text-field(
```
subdavis: @matthewma7 I've found that moving the v-text-field into the activator slot has the undesired effect of bringing up the "no results" popover menu as soon as a user clicks into the text box.

I don't really like showing the user error messages for the initial state.  We could either stick a v-if on the v-list below, or move v-text-field back out of this slot, which would require use of `input-activator`.

I don't know which would be better.

matthewma7: Hm, there is an` open-on-click` prop that we can use. 

But overall the reason I meant to use the activator slot is that I think we are trying to use this library as a standard and example for other projects, so I think we could use as much standard stuff as possible. Otherwise, people (like I was) will have questions on why we didn't standard activator because the behavior of full-size input and non-open menu are not explicit and got confused.

subdavis: I certainly agree that we should use standard vuetify features when possible, as long as they make sense to users. 

`open-on-click` works perfectly.  Thank you!

subdavis:
screenshot_20190207_141301

Most recent commit adds "Skeleton Results" to search results while real results are loading.  I decided to give this a try after realizing that:

1. On poor connections, the previous behavior was to show "No results for...." while the search query loads.  
2. Replacing that with a loading spinner or progress bar didn't look nice.

The "text" bars for the skeleton results shimmer as you would expect. 

Loading is debounced so those results don't flash on regular queries.  To see them, either point GWC at DKC or else use Chrome's 3G network throttling.

subdavis:
Closing in favor of #81

PR #61, opened 2019-01-21
Search improvements

matthewma7:
Hi @subdavis,

This the PR I mentioned in https://github.com/girder/girder_web_components/pull/59. Do you think they make sense?


src/components/Search.vue
```diff
@@ -1,14 +1,12 @@
...
-v-layout.searchbar(row, justify-center, align-center)
+v-layout.girder-searchbar(row, align-center)
   v-icon.mdi-24px.mx-2(color="white") {{ $vuetify.icons.search }}
-  v-text-field.searchtext.mx-4(
-      light, solo, hide-details, clearable,
-      v-model="searchtext")
-  v-menu.searchbar(
-      offset-y, max-width,
-      :activator="'.searchtext'",
+  v-menu.searchbar.grow(
```
subdavis: Please add `content-class="girder-searchbar"` to this v-menu and remove `.searchbar`.  The searchbar style has no effect.

src/components/Search.vue
```diff
@@ -22,21 +20,20 @@ v-layout.searchbar(row, justify-center, align-center)
           v-list-tile-title No results found for query '{{ searchtext }}'
           v-list-tile-sub-title Try an advanced search or refine your query.
       v-list-tile(v-else-if="showMore && searchresults.length > maxQuickResults",
-          @click="$emit('more-results', searchtext)")
+          @click="$emit('more-results', searchtext)",
+          :disabled="!$listeners['more-result']")
```
subdavis: great addition.  You don't see $listeners used much.

src/components/Search.vue
```diff
@@ -132,4 +127,28 @@ export default {
     }
   }
 }
+
+.girder-arrow-menu {
```
subdavis: How can we refactor this class for use elsewhere in Girder Web Components?

I created MenuActivatorBtn because I wanted to use this type of popover style elsewhere.  This isn't tightly coupled to search, so I think it might be worth investigating how to have some global style.

If we decide that this style isn't abstract enough to refactor globally, could we rename it `.girder-search-arrow-menu`.  I personally think it can be refactored.

src/components/Search.vue
```diff
@@ -22,21 +20,20 @@ v-layout.searchbar(row, justify-center, align-center)
...
           v-list-tile-sub-title Try an advanced search or refine your query.
       v-list-tile(v-else-if="showMore && searchresults.length > maxQuickResults",
-          @click="$emit('more-results', searchtext)")
+          @click="$emit('more-results', searchtext)",
+          :disabled="!$listeners['more-result']")
         v-list-tile-action
           v-icon {{ $vuetify.icons.more }}
         v-list-tile-content
           v-list-tile-title more
   v-menu.searchbar(
```
subdavis: This class also has no effect.  I can remove it in my own PR once this is merged.

src/components/Search.vue
matthewma7: There were three .searchbar from the beginning, I later renamed the top level one. I am thinking to remove the .searchbar off the two v-menu, will that functionally work?

src/components/Search.vue
matthewma7: I learned about this when I see Vuetify component behave differently when based on if there is a listener. So, I think this is safe to do. 🙂

src/components/Search.vue
matthewma7: Renaming is a good idea.
This girder-arrow-menu is not far from being able to be used globally, but it is still a missing option to be aligned left. I think when there is another use case for the arrow button, we could revisit and consider making it a global utility then.

subdavis:
LGTM.  Can I merge?

PR #63, opened 2019-01-21
Add useful constants to the library

zachmullen:
The second commit also adds a mixin. I'm happy to separate these commits into different PRs if people prefer.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,0 +1,12 @@
+export const AccessType = {
+  NONE: -1,
+  READ: 0,
+  WRITE: 1,
+  ADMIN: 2,
+  SITE_ADMIN: 100,
+};
+
+export const SortDir = {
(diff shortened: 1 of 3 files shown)


subdavis:
I don't think different PRs are necessary, myself.  These will come in handy when we start working on a details widget.   I think we should revisit the design @jtomeck drew up referenced in #31 

Do you have any general guidance about what belongs in constants?  Is it mostly just numeric enums that correspond to the girder data models?  For example,  should `_modelType` values be refactored into constants?

Looks good to me.

zachmullen:
[Here's the constants.js from the Backbone client](https://github.com/girder/girder/blob/master/girder/web_client/src/constants.js). Basically any set of constants that are meant to be used across multiple modules would be good candidates to put here.

PR #69, opened 2019-01-25
Handle admin approval registration

matthewma7:
This PR handles **admin approval required** and/or **email verification required** registration by simply give a generic message.
This PR also address two related small issues and added a minimal test to the Register component.

### DIFF
--- a/src/components/Authentication/Register.vue
+++ b/src/components/Authentication/Register.vue
@@ -7,8 +7,15 @@
       :key="err",
       :value="true",
       type="error") {{ err }}
+  v-alert.mt-0(
+      dismissible,
+      transition="scale-transition",
+      v-for="info in alerts.infos",
+      :key="info",
+      :value="true",
+      type="info") {{ info }}
   v-container
-    v-form(@submit.prevent="register", ref="register")
+    v-form(@submit.prevent="register", ref="form")
       v-text-field(
           v-model="login",
           label="Username",
@@ -79,6 +86,7 @@ export default {
       inProgress: false,
       alerts: {
         errors: [],
+        infos: [],
       },
       nonEmptyRules,
       retypeMustMatchPasswordRules,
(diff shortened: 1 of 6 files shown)


src/components/Authentication/Register.vue
```diff
@@ -7,8 +7,15 @@
       :key="err",
       :value="true",
       type="error") {{ err }}
+  v-alert.mt-0(
+      dismissible,
+      transition="scale-transition",
+      v-for="info in alerts.infos",
+      :key="info",
+      :value="true",
+      type="info") {{ info }}
```
subdavis: Small nitpick: can we color this as primary?  There's a slight shade difference here.

screenshot_20190128_092335

matthewma7: I think some difference could make it stand out more. Generally, I am thinking since Vuetify has built-in types for v-alert, we should use it semantically correctly because colors are customizable in sub-applications. Following Vuetify would make GWC fit in sub-application better. Does this sound reasonable?

subdavis: I think an intentional difference, like having a unique `alert:info` color, would be fine, but these shades are so close that I don't think it stands out so much as looks like a mistake.  Vuetify's behavior is to use the `info` color for info alerts, which is by default the same as the primary color.  

I suppose a more correct way to fix this would be to override `info` in `utils/vuetifyConfig.js` to be the same as our overridden primary.  Girder Web Components already makes you override that config in order to change colors and themes.  Could we do that?

matthewma7: By default, Vuetify has different colors for primary and info. It a set of two colors has simlar hue. It actually looks so bad. I could change info to a color has similar hue of the customized primrary. BTW: What is the reason we customized default Vuetify's color?

Default:
image

Changed info:
(Same Hue as primary but lighter)
image

subdavis: Oh yeah that does look weird.  Alright, let's not worry about the color issue here.  There may be a larger discussion on color to be had.

I'm happy to merge this as-is if that's ok.

matthewma7: Cool, could you approve this PR again? (Becuase I pushed again after you approved it). Thank you for reviewing this PR. 🙂

PR #70, opened 2019-01-25
Circleci/deploy

### LINKED ISSUE #68: Add Github Pages deployment to CI
I think it would be nice if we could deploy the example app to GH pages so that other folks in D&A can poke around and give feedback/suggestions.  It's also generally good practice for communication.

This should be a pretty easy addition.  I'm happy to do it if this sounds valuable.

subdavis:
PR open to test CI runner.

Requires updates to @vue/cli-service and the eslint tool to enable `publicPath` support in vue config.

fixes #68

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,6 +57,3 @@ workflows:
       - deploy:
           requires:
             - test
-          filters:
-            branches:
-                only: master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,3 +57,6 @@ workflows:
       - deploy:
           requires:
             - test
+          filters:
+            branches:
+                only: master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2
 jobs:
-  build:
+  checkout:
     docker:
       - image: circleci/node:10
     steps:
@@ -15,9 +15,48 @@ jobs:
           key: yarn-packages-{{ checksum "yarn.lock" }}
           paths:
             - ~/.cache/yarn
+      - persist_to_workspace:
+          root: /home/circleci
+          paths:
+            - project
+
+  test:
+    docker:
+      - image: circleci/node:10
+    steps:
+      - attach_workspace:
+          at: /home/circleci
       - run: yarn lint --no-fix
       - run: yarn lint:pug
       - run: yarn lint:style
       - run: yarn test:unit --coverage --coverageReporters html
       - store_artifacts:
           path: coverage/
+
+  deploy:
+    docker:
+      - image: circleci/node:10
+    steps:
+      - attach_workspace:
+          at: /home/circleci
+      - run: yarn build:demo
+      - add_ssh_keys:
+          fingerprints:
+            - "eb:dc:cd:0a:f4:99:56:3f:79:f7:22:a2:6c:21:56:c7"
+      - deploy:
+          command: .circleci/pages-deploy.sh
+
+workflows:
+  version: 2
+  checkout_test_deploy:
+    jobs:
+      - checkout
+      - test:
+          requires:
+            - checkout
+      - deploy:
+          requires:
+            - test
+          filters:
+            branches:
+                only: master
--- a/.circleci/pages-deploy.sh
+++ b/.circleci/pages-deploy.sh
@@ -1,0 +1,18 @@
+git config user.name "$USER_NAME"
+git config user.email "$USER_EMAIL"
+
+# Remove strict host checking for github.com
+mkdir -p ~/.ssh/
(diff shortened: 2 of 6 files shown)


.circleci/config.yml
```diff
@@ -15,9 +15,48 @@ jobs:
           key: yarn-packages-{{ checksum "yarn.lock" }}
           paths:
             - ~/.cache/yarn
+      - persist_to_workspace:
+          root: /home/circleci
+          paths:
+            - project
```
subdavis: Not sure if there's a more conventional way of saying "persist the whole workspace".

.circleci/config.yml
```diff
@@ -15,9 +15,48 @@ jobs:
...
+      - image: circleci/node:10
+    steps:
+      - attach_workspace:
+          at: /home/circleci
+      - run: yarn build:demo
+      - add_ssh_keys:
+          fingerprints:
+            - "eb:dc:cd:0a:f4:99:56:3f:79:f7:22:a2:6c:21:56:c7"
+      - deploy:
+          command: .circleci/pages-deploy.sh
```
subdavis: This job only runs on master branch.  However, it might be good to double-check this with bash's `if [ $env == 'master' ] ...`

I don't think that's necessary for something like this though.

.circleci/pages-deploy.sh
```diff
@@ -0,0 +1,18 @@
+git config user.name "$USER_NAME"
+git config user.email "$USER_EMAIL"
```
subdavis: I have these variables set as "Girder" and "girder-developers@kitware.com" respectively

.circleci/pages-deploy.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# Remove strict host checking for github.com
+mkdir -p ~/.ssh/
+echo -e "Host github.com\n\tStrictHostKeyChecking no\n\tUserKnownHostsFile /dev/null\n" >> ~/.ssh/config
+chmod 600 ~/.ssh/config
```
subdavis: It confuses me that this is necessary.  You'd think CircleCI could disable strict host checking for custom keys where domain is set.   It prevents the feature of being able to add your own private ssh keys unusable without this.

src/main.js
```diff
@@ -4,7 +4,11 @@ import App from './App.vue';
 import RestClient from './rest';
 
 Vue.use(Girder);
-const girderRest = new RestClient({ apiRoot: 'http://localhost:8080/api/v1' });
+const girderRest = new RestClient({
+  apiRoot: process.env.NODE_ENV === 'production'
+    ? 'https://data.kitware.com/api/v1'
+    : 'http://localhost:8080/api/v1',
+});
```
subdavis: We could be more rigorous about this with `DefinePlugin` but this seems fine for a demo site.

subdavis:
Looks like this is good to go.  I disabled branch checking and was able to deploy this.  Let me know how this looks.

I'm new to CircleCI so I leaned heavily on examples/docs.  I generally really like how straightforward the workflow management is.

subdavis:
@zachmullen @matthewma7 

Are there any changes required here?  Happy to follow up wherever is needed.   Thanks again.

yarn.lock
```diff
@@ -40,20 +40,20 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
-"@babel/core@^7.0.0":
```
zachmullen: This lockfile got changed but our dependencies were not changed on this branch. Is there a reason to include these changes on this PR?

zachmullen: Nevermind, just saw your comment about publicPath support

vue.config.js
```diff
@@ -1,3 +1,6 @@
 module.exports = {
   lintOnSave: false,
+  publicPath: process.env.NODE_ENV === 'production'
+    ? '/girder_web_components'
```
zachmullen: This means that when doing `yarn build` to build the dist package, our `publicPath` will get set to `/girder_web_components`. Is that OK? What affect will that have on the bundle for downstreams (i.e. environments other than gh-pages)?

subdavis: I don't believe publicPath affects builds in `--target lib` mode.  

https://github.com/vuejs/vue-cli/issues/2196 -- looks like publicPath is overridden.

zachmullen: Good to know. One thing that might be worth doing is putting a comment here on line 4 that says "this should only affect gh-pages deployment" so we know why it's here.

subdavis:
@zachmullen can you force-merge?  That pending CI job was manually killed because it would have run forever.  

CI was waiting on user input to accept remote fingerprint, solved by adding line 6 in `.circleci/pages-deploy.sh`.

PR #72, opened 2019-02-07
Circleci/deploy2

subdavis:
Duplicate of #70 because webhooks failed.

Edit: that didn't work at all as I intended

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2
 jobs:
-  checkout:
+  test:
     docker:
       - image: circleci/node:10
     steps:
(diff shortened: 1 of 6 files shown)


zachmullen:
It seems like we need to change the expected status checks

zachmullen:
What's the rationale for separating this into 3 separate CircleCI jobs? Since the dependency graph is just a line (no parallelism), would it make more sense to simply combine them as steps of the same job?

subdavis:
Fixed the expected status checks.

We could definitely combine the first two jobs. 

I used branch filtering to prevent deployments on anything except the master branch -- this seems like the preferred way to accomplish that with CircleCI.

Ideally we would also use CircleCI to deploy the UMD library, which could run parallel to a GH Pages deployment.

subdavis:
Alright, combined the first 2.

PR #75, opened 2019-02-08
Fix styles on data browser component

zachmullen:
@subdavis PTAL, I couldn't determine what effect the `elevation-1` class was intended to have on this component, removing it didn't change anything that I could find. If there's a reason for it I will put it back.

### DIFF
--- a/src/components/DataBrowser.vue
+++ b/src/components/DataBrowser.vue
@@ -162,7 +162,7 @@ export default {
 </script>
 
 <template lang="pug">
-v-data-table.girder-file-browser-component.elevation-1(
+v-data-table.girder-file-browser-component(
     select-all,
     :headers-length="4",
     v-model="selected",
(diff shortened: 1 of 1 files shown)


src/components/DataBrowser.vue
```diff
@@ -257,8 +255,12 @@ v-data-table.girder-file-browser-component.elevation-1(
         border-right: 1.5px solid;
       }
 
-      .text-container i {
-        vertical-align: bottom;
+      .text-container {
+        vertical-align: text-bottom;
```
subdavis: Except for the addition of this style.

zachmullen:
In that case, is this good to merge?

zachmullen:
Or do you just want to incorporate the alignment fix onto your topic?

subdavis:
Got a bit sidetracked.  Sure, we can merge this.  

If possible, could we start reviewing the search branch to get that merged?  I know it's pretty big, so I can try to break some of the pieces apart.

zachmullen:
Sure thing, if it's sensibly breakable that'd be great.

PR #76, opened 2019-02-08
Range select in Data Browser

zachmullen:
This is based on top of #75.

### DIFF
--- a/src/components/Presentation/DataTable.vue
+++ b/src/components/Presentation/DataTable.vue
@@ -26,6 +26,23 @@ export default {
       required: true,
     },
   },
+  data() {
+    return { lastCheckBoxIdx: null };
+  },
+  methods: {
+    handleRowSelect({ shiftKey }, props) {
+      // eslint-disable-next-line no-param-reassign
+      props.selected = this.selectEnabled ? !props.selected : false;
+      if (this.selectEnabled && shiftKey && this.lastCheckBoxIdx !== null) {
+        const [start, end] = [this.lastCheckBoxIdx, props.index + 1].sort();
+        const newlySelectedRows = this.rows
+          .slice(start, end)
+          .filter(row => this.value.find(el => el._id === row._id) === undefined);
+        this.$emit('input', newlySelectedRows.concat(this.value));
+      }
+      this.lastCheckBoxIdx = props.index;
+    },
+  },
 };
 </script>
 
@@ -40,14 +57,18 @@ v-data-table.girder-data-table(
     :items="rows",
     :total-items="totalItems",
     :loading="loading ? 'accent' : false",
-    item-key="_id")
+    item-key="_id",
+    ref="girderDataTable")
 
   template(slot="headers", slot-scope="props")
-    slot(name="header", :props="props")
+    slot(name="header",
+        :all="props.all",
+        :indeterminate="props.indeterminate",
+        :headers="props.headers")
 
   template(slot="items", slot-scope="props")
     tr.itemRow(:active="props.selected",
-        @click="if (selectEnabled) props.selected = !props.selected;",
+        @click="handleRowSelect($event, props)",
         :key="props.index")
       td.pl-3.pr-0(v-if="selectEnabled")
         v-checkbox.secondary--text.text--darken-1.pr-2(
--- a/src/components/Presentation/DataTable.vue
+++ b/src/components/Presentation/DataTable.vue
@@ -57,8 +57,7 @@ v-data-table.girder-data-table(
     :items="rows",
     :total-items="totalItems",
     :loading="loading ? 'accent' : false",
-    item-key="_id",
-    ref="girderDataTable")
+    item-key="_id")
 
   template(slot="headers", slot-scope="props")
     slot(name="header",
--- a/src/components/Presentation/DataTable.vue
+++ b/src/components/Presentation/DataTable.vue
@@ -31,7 +31,6 @@ export default {
   },
   methods: {
     handleRowSelect({ shiftKey }, props) {
-      // eslint-disable-next-line no-param-reassign
       props.selected = this.selectEnabled ? !props.selected : false;
       if (this.selectEnabled && shiftKey && this.lastCheckBoxIdx !== null) {
         const [start, end] = [this.lastCheckBoxIdx, props.index + 1].sort();
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
       "@vue/airbnb"
     ],
     "rules": {
+      "no-param-reassign": ["error", { "props": false }],
       "no-underscore-dangle": 0,
       "import/prefer-default-export": 0
     },


src/components/DataBrowser.vue
```diff
@@ -147,8 +159,9 @@ export default {
         promises.push(this.girderRest.get(GIRDER_ITEM_ENDPOINT, { params: itemParams }));
       }
       const responses = (await Promise.all(promises)).map(response => response.data);
-      const rows = [].concat.apply(...responses).map(item => ({
+      const rows = responses.flat().map(item => ({
```
zachmullen: This change is just for readability, but the downside is that `Array.prototype.flat` only exists in node>=11, and is not supported on Edge or IE, and I'm unsure of exactly what our polyfill situation is. I can revert this if it's breaking compatibility with environments that we care about.

subdavis: I actually ran into this very recently.  I used flat() on some of my video component stuff and found that it was only recently supported in chrome (version 69  -- the last version with more angled tabs). The current babel config won't polyfill for this, but I didn't look into what would.

subdavis:
Could we hold off on further changes to data browser? 

A rebase on this branch will be much easier than one on search.  I can rebase this one after we get the others in.

As I mentioned in #75, if I need to break search into several smaller PRs I can do that.

zachmullen:
Sure. I'm going to rebase this and remove the extraneous changes, so that it will become quite small, then we can sit on this PR until the other is merged.

zachmullen:
@subdavis now that #79 is merged, do you have time to integrate this feature? I took a stab at it but it wasn't as straightforward now that the components are separated.

subdavis:
@zachmullen yep, going to get this and #88 squared away.

subdavis:
This was actually quite an interesting task.  I was able to do it while still allowing `v-data-table` to maintain selection state the way it wants, but it was tricky.

src/components/Presentation/DataTable.vue
```diff
@@ -26,6 +26,23 @@ export default {
       required: true,
     },
   },
+  data() {
+    return { lastCheckBoxIdx: null };
+  },
+  methods: {
+    handleRowSelect({ shiftKey }, props) {
+      // eslint-disable-next-line no-param-reassign
```
zachmullen: In a previous incarnation of this branch I had advocated for disabling this rule globally via:

```json
no-param-reassign: ["error", { "props": false }]
```

How do people feel about that? @subdavis @jbeezley @matthewma7

zachmullen:
One of the things I did on the old branch was to remove a manual synchronization between the `selected` property of each row, and the `selected` Array in the data section of the DataBrowser. Is there a way to consolidate that state in the new design as well?

src/components/Presentation/DataTable.vue
subdavis: I personally like that rule and think it should be followed whenever possible.  We only broke it here because vuetify implements the `toggle()` behavior through a setter function on `props.selected`.

zachmullen: I think the rule is *weird*, because it has two (seemingly) unrelated parts that are lumped into the same top-level rule.

1. No reassignment of params.
2. No *modification* of params.

The second case is there to promote more functional-style programming by disallowing side-effects on input params. The first one has nothing to do with side effects, and is in a completely different category of potential problems. I think the first one is much more in the realm of *linting* and the second is much more about *programming patterns*.

I feel comfortable programming with either of those things disabled, but if others think it might actually catch us doing unintended or bad things, we can keep the default and I can sprinkle in line-level disables.

subdavis:
@zachmullen What do you see as the drawbacks of separating state into a separate variable?  Your branch short-circuited the selection management code inside `v-data-table` and kept state as a property of row.

Both approaches seem pretty harmless to me, but I changed it to make `DataTable.vue` follow the same pattern as `v-data-table` and to prevent the `row.selected` property from being leaked out to upstream when `DataBrowser` emits `itemclick` and `update:location`.

I'm fine with changing it back, but I'm interested in why you preferred to consolidate them.

zachmullen:
It seemed like we were keeping the list of selected rows in two places rather than having a single source of truth. The change I made was an attempt to fix that by moving `selected` out of `data` and into `computed`, as

```javascript
selected() {
  return this.rows.filter(r => r.selected);
}
```

We would ideally either keep selected as a list of rows, or keep a `selected` boolean property on each row, but not both places.

subdavis:
It is my intention that the source of truth is the selected array in DataBrowser's data.

Where is the other?

zachmullen:
> Where is the other?

The references to `props.selected` in DataTable.vue

src/components/Presentation/DataTable.vue
matthewma7: I found this one a little bit weird because it prevented me form doing `Function.prototype={}`

subdavis: Alright, Zach and Matthew have made a good case.  Let's drop this rule.

subdavis:
> The references to `props.selected` in DataTable.vue

I don't consider this a second source of truth.  `props.selected` is the `v-data-iterator` way of providing `v-model` capability to a checkbox or other user input device inside the slot.

Its implementation makes it [a computed property](https://github.com/vuetifyjs/vuetify/blob/33f403342d3eccbd1a445c4c676bf8b02bce8303/packages/vuetify/src/mixins/data-iterable.js#L216) just like the `selected` computed prop from your branch -- it is generated as part of `v-data-table's` render function.

## an example

I think of `v-data-table` as just a super-complicated wrapper around [checkbox array](https://vuetifyjs.com/en/components/selection-controls#checkboxes-array).

If I were to create a new component type `v-checkbox-group` that worked like `v-radio-group`, it would look like this:

```javascript
<template lang="pug">
.my-form
  v-checkbox-group(v-model="selected")
</template>

<script>
export default {
  data() {
    return { selected: [ true, false, true ] };
  }
}
</script>
```

There's only 1 source of truth here.

If I want to extend `v-checkbox-group` with a slot so the consumer could decorate their checkboxes (or whatever):

```javascript
<template lang="pug">
.my-form
  v-checkbox-group(v-model="selected")
    template(slot="checkbox", slot-scope="props")
      v-checkbox(v-model="props.selected")
</template>

<script>
export default {
  data() {
    return { selected: [ true, false, true ] };
  }
}
</script>
```

`props.selected` is a little misleading, because under the hood, it [is just a computed value and a handler on the settter](https://github.com/vuetifyjs/vuetify/blob/33f403342d3eccbd1a445c4c676bf8b02bce8303/packages/vuetify/src/mixins/data-iterable.js#L337).

When the setter for selected is called, `v-data-table` emits an `input` event.  We shouldn't have to worry about these details anyway, because `v-data-table` guarantees the `v-model` interface.  Our code isn't doing any state syncing, and their code can do whatever it wants.

## differences

I just realized there was a behavioral difference between our two branches, which I think is more important than the implementation discussion above.

When selected state is moved into a separate array, it can persist through pagination and navigation, and we have control over whether or not it it actually does.  Moving between paginated results in the data browser loses whatever was selected in the last page if we consolidate that state.

If we want to use data browser for bulk operations on items within a folder, or even if the option should be available to upstream, I think it makes more sense not to consolidate.

zachmullen:
Discussed with @subdavis , turns out the `props.selected` is actually computed, so there is no duplication of state. This is an idiosyncrasy of Vuetify.

This LGTM.

PR #79, opened 2019-02-14
Split datatable and browser into separate components

subdavis:


zachmullen:
I went back to #59 to try and discover the rationale for this change, I found

> * Refactor data table from DataBrowser into its own component so that it can be used to render any data list from girder, especially search results.

but #81 doesn't use this feature. Is it used in some other display of search results?

(Just curious, the change seems fine generally.)

zachmullen:
I just want to understand the mission of this PR, do we anticipate using this DataTable in some upcoming components?

matthewma7:
I don't know what @subdavis planned for DataTable. @zachmullen I understand if you think this increase API surface, but I personally think extracting the DataTable could be a good thing in downstream apps because it doesn't have the header bar (breadcrumb, new folder), it's less intrusive and would aesthetically fit better.

zachmullen:
> I understand if you think this increase API surface

It actually doesn't seem to be part of the library API, just the internal one. My concern wasn't that, it was simply that it might be premature separation, but since it's already done and is a good idea, we can go ahead and merge it.

zachmullen:
Merging so I can put #76 on top of this

PR #80, opened 2019-02-14
fix for building girder web components from external app

subdavis:
Other apps can't resolve '@'

### DIFF
--- a/src/utils/mixins.js
+++ b/src/utils/mixins.js
@@ -1,4 +1,4 @@
-import { AccessType } from '@/constants';
+import { AccessType } from '../constants';
 
 /**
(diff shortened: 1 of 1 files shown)


zachmullen:
Good catch, is there a way we can catch this with our linter?

subdavis:
Maybe this?

https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unresolved.md#importno-unresolved

zachmullen:
Thanks for the link, we can merge this and I'll look into a linter solution.

subdavis:
I installed that plugin and tried a few things, but they all allowed the @ import.

PR #81, opened 2019-02-14
Searchbar

subdavis:
Add searchbar.  Replaces #59.  See discussion on that PR.

### DIFF
--- a/src/components/Search.vue
+++ b/src/components/Search.vue
@@ -145,23 +145,9 @@ v-layout.girder-searchbar(row, align-center)
   .skeleton.skeleton--text {
     background: linear-gradient(270deg, #e0e0e0, #c7c7c7, #e0e0e0);
     background-size: 600% 600%;
-    -webkit-animation: AnimationName 2s ease infinite;
-    -moz-animation: AnimationName 2s ease infinite;
-    animation: AnimationName 2s ease infinite;
+    animation: SkeletonShimmer 2s ease infinite;
 
-    @-webkit-keyframes AnimationName {
-      0% { background-position: 0% 51%; }
-      50% { background-position: 100% 50%; }
-      100% { background-position: 0% 51%; }
-    }
-
-    @-moz-keyframes AnimationName {
-      0% { background-position: 0% 51%; }
-      50% { background-position: 100% 50%; }
-      100% { background-position: 0% 51%; }
-    }
-
-    @keyframes AnimationName {
+    @keyframes SkeletonShimmer {
       0% { background-position: 0% 51%; }
       50% { background-position: 100% 50%; }
       100% { background-position: 0% 51%; }
--- a/src/components/Search.vue
+++ b/src/components/Search.vue
@@ -75,7 +75,7 @@ v-layout.girder-searchbar(row, align-center)
   v-icon.mdi-24px(color="white") {{ $vuetify.icons.search }}
   v-menu.grow.mx-3(
       offset-y, content-class="girder-searchbar-menu", :open-on-click="false",
-      :value="searchText", :nudge-bottom="6")
+      :value="searchText", :nudge-bottom="6", transition="slide-y-transition")
     v-text-field(
         slot="activator", v-model="searchText", light, solo, hide-details, clearable)
     v-list
--- a/src/components/Search.vue
+++ b/src/components/Search.vue
@@ -1,9 +1,10 @@
 
 <script>
-import { countlock } from '../utils/mixins';
+import { DebounceCounter } from '../utils';
+
+const counter = new DebounceCounter();
 
 export default {
-  mixins: [countlock],
   props: {
     maxQuickResults: {
       type: Number,
@@ -25,7 +26,7 @@ export default {
   },
   computed: {
     loading() {
-      return this.locked;
+      return counter.flag;
     },
     quickResults() {
       return this.searchResults.slice(0, this.maxQuickResults);
@@ -45,7 +46,7 @@ export default {
       default: [],
       async get() {
         let results = [];
-        this.lock();
+        counter.inc();
         try {
           if (this.searchText) {
             const { data } = await this.girderRest.get('resource/search', {
@@ -56,7 +57,7 @@ export default {
         } catch (err) {
           this.$emit('error', err.message || 'Unknown error during search');
         }
-        this.unlock();
+        counter.dec();
         return results;
       },
     },
@@ -100,7 +101,10 @@ v-layout.girder-searchbar(row, align-center)
         v-list-tile-content
           v-list-tile-title More
       //- Skeleton search results shown as "loading" animation
-      v-list-tile(v-show="loading", v-for="i in (maxQuickResults + (showMore ? 1 : 0))")
+      v-list-tile(
+          v-show="loading",
+          v-for="i in (maxQuickResults + (showMore ? 1 : 0))",
+          :key="`skeleton-${i}`")
         v-list-tile-action
           v-icon.grey--text.text--lighten-1 {{ $vuetify.icons.circle }}
         v-list-tile-content
@@ -129,8 +133,14 @@ v-layout.girder-searchbar(row, align-center)
 </template>
 
 <style lang="scss">
-.girder-searchbar .v-text-field .v-input__control {
-  min-height: 40px;
+.girder-searchbar {
+  .v-text-field .v-input__control {
+    min-height: 40px;
+  }
+
+  .v-menu__activator * {
+    cursor: text !important;
+  }
 }
 
 .girder-searchbar-menu {
--- a/src/components/Search.vue
+++ b/src/components/Search.vue
@@ -1,0 +1,196 @@
+
+<script>
+import { countlock } from '../utils/mixins';
+
+export default {
+  mixins: [countlock],
+  props: {
+    maxQuickResults: {
+      type: Number,
+      default: 6,
+    },
+    showMore: {
+      type: Boolean,
+      default: true,
+    },
+  },
+  inject: ['girderRest'],
+  data() {
+    return {
+      searchText: '',
+      searchMode: 'prefix',
+      searchTypes: ['user', 'folder', 'item'],
+      searchOptionsMenu: false,
+    };
+  },
+  computed: {
+    loading() {
+      return this.locked;
+    },
+    quickResults() {
+      return this.searchResults.slice(0, this.maxQuickResults);
+    },
+    searchParams() {
+      return {
+        q: this.searchText,
+        mode: this.searchMode,
+        types: JSON.stringify(this.searchTypes),
+        // + 1 to determine if total results > maxQuickResults
+        limit: this.maxQuickResults + 1,
+      };
+    },
+  },
+  asyncComputed: {
+    searchResults: {
+      default: [],
+      async get() {
+        let results = [];
+        this.lock();
+        try {
+          if (this.searchText) {
+            const { data } = await this.girderRest.get('resource/search', {
+              params: this.searchParams,
+            });
+            results = [].concat(...this.searchTypes.map(t => data[t]));
+          }
+        } catch (err) {
+          this.$emit('error', err.message || 'Unknown error during search');
+        }
+        this.unlock();
+        return results;
+      },
+    },
+  },
+  methods: {
+    formatUsername(user) {
+      return `${user.firstName} ${user.lastName} (${user.login})`;
+    },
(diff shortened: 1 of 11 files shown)


src/utils/mixins.js
```diff
@@ -32,4 +32,48 @@ const accessLevelChecker = {
   },
 };
 
-export { accessLevelChecker, sizeFormatter };
+/**
+ * Provides read/write-style lock with non-reactive components such that
+ * the lock can be used as a side-effect in (especially async) computed properties.
+ * Includes debouncing.
```
zachmullen: It took me a while to understand exactly why this was being used. I think part of that is due to the name, "lock" is a little misleading IMO, this is really a non-reactive counter (intentionally designed to circumvent reactivity on the counter itself) that debounces the setting of a reactive flag. Even in this use case, it's not really "locking" anything.

I can't decide whether or not I think this is too early to generalize into a mixin, do you have a strong feeling one way or the other? (Related, I think `formatUserName` might actually be a good candidate for a mixin...)

src/components/Search.vue
```diff
@@ -0,0 +1,196 @@
...
+      default: [],
+      async get() {
+        let results = [];
+        this.lock();
+        try {
+          if (this.searchText) {
+            const { data } = await this.girderRest.get('resource/search', {
+              params: this.searchParams,
+            });
+            results = [].concat(...this.searchTypes.map(t => data[t]));
```
zachmullen: We aren't guaranteed to receive these responses in the order that we sent the requests, i.e. our results might not be for the value currently typed into the search box. Should we do anything to try to resolve that? (Feel free to file an issue, we don't necessarily need to resolve this on this PR.)

src/components/Search.vue
```diff
@@ -0,0 +1,196 @@
...
+</script>
+
+<template lang="pug">
+v-layout.girder-searchbar(row, align-center)
+  v-icon.mdi-24px(color="white") {{ $vuetify.icons.search }}
+  v-menu.grow.mx-3(
+      offset-y, content-class="girder-searchbar-menu", :open-on-click="false",
+      :value="searchText", :nudge-bottom="6")
+    v-text-field(
+        slot="activator", v-model="searchText", light, solo, hide-details, clearable)
```
zachmullen: I'm seeing a pointer cursor on the search text field, tracked it down to the following vuetify rule:

```css
.v-menu__activator * {
    cursor: pointer;
}
```

Is it a bad idea to use a `v-text-field` as an activator for a `v-menu`? Is it easy to do this some other way? If not, we can proceed with this and just override the CSS...

src/components/Search.vue
```diff
@@ -0,0 +1,196 @@
...
+    -moz-animation: AnimationName 2s ease infinite;
+    animation: AnimationName 2s ease infinite;
+
+    @-webkit-keyframes AnimationName {
+      0% { background-position: 0% 51%; }
+      50% { background-position: 100% 50%; }
+      100% { background-position: 0% 51%; }
+    }
+
+    @-moz-keyframes AnimationName {
```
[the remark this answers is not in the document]
subdavis: These are unique to the skeleton shimmer animation.  I can/should probably name it something more useful.

I don't know what you mean by inserting them automatically.  These rules are just like any other custom CSS right?  I tried removing them and the animation disappeared.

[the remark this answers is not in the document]
zachmullen: The [autoprefixer](https://github.com/postcss/autoprefixer) postcss plugin that we use is supposed to automatically insert the browser-specific versions of rules automatically, so you shouldn't (usually) need to insert the `-moz`, `-webkit` versions of rules.

src/components/Search.vue
```diff
@@ -0,0 +1,196 @@
...
+      //- Skeleton search results shown as "loading" animation
+      v-list-tile(v-show="loading", v-for="i in (maxQuickResults + (showMore ? 1 : 0))")
+        v-list-tile-action
+          v-icon.grey--text.text--lighten-1 {{ $vuetify.icons.circle }}
+        v-list-tile-content
+          v-list-tile-title.skeleton.skeleton--text.mb-2(
+              :style="{ width: (60 + (4 * (i % 3))) + '%', height: '10px' }")
+          v-list-tile-sub-title.skeleton.skeleton--text(
+              :style="{ width: (45 - (4 * (i % 2))) + '%', height: '6px' }")
+  v-menu(
```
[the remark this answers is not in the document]
subdavis: I agree they aren't associated well.  I suppose some of the design in #40 doesn't apply now that the searchbar isn't full width and can appear next to other content.  That being said, I've never seen the attachment style used in Girder UI in a material-design application.  It's distinctly [bootstrap 3](https://getbootstrap.com/docs/3.3/components/#input-groups) and would look out of place here, I think.   

Now that I think about it, I can't come up with a similar example application where query refinement tools are provided on a quick-search bar.  Those tools always appear on their own advanced search page.  

The only thing I can think of to do is replace the dismiss button `X` with this button.  That dismiss is somewhat pointless anyway.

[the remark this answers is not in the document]
zachmullen: # 🤷‍♂️

paging Dr. Tomeck

src/utils/mixins.js
subdavis: The mixin is so-named because it's basically the read- part of a reader-writer-lock.  The debounce part does detract from that though, so it's not usable as an actual resource lock.

The behavior of the mixin is generic enough to separate.  Its only conceivable purpose is handling render of busy spinners and other graphics.  I don't feel that strongly about it though.   If we find it can be recycled, we can refactor it then.

I would propose:
*  renaming it to debounceCounter, with methods `inc()` and `dec()`, and a Boolean data property `isZero`.

Otherwise, pulling it into search with the same method names.

src/components/Search.vue
subdavis: Yeah, that's true...    

Let's log an issue about it and see how it relates to #83 ?

src/utils/mixins.js
zachmullen: I do like the name `debounceCounter` better. Just throwing stuff out there without fully thinking it through, what if we did it as a self-contained class rather than a mixin?

```javascript
class DebounceCounter extends Vue {
    constructor(delay=200) {
        super({
          data: {
            flag: false,
          },
        });
        this.delay = 200;
        this.count = 0;
    }

    inc() ...
    dec() ...
    _toggleFlag(val) ...
```

That encapsulates things nicely behind an abstraction rather than leaking several names into the components that use it. @jbeezley @matthewma7 @subdavis any thoughts? Better suggestions?

zachmullen: (It also has the benefit of allowing a component to potentially use multiple debouncing counters that are independent, which the mixin would not)

src/components/Search.vue
zachmullen: Sounds good to me. I was thinking we could just keep track of the last Promise that was issued and throw out any that we receive that aren't from that one. Will open the issue.

matthewma7: I don't think Vuex has utilities to handle this kind of issue. Keep a queue like @zachmullen would work good enough for just this case, but this kind issue generally is asking for things like https://github.com/vuejs/vue-rx

zachmullen: #85

src/components/Search.vue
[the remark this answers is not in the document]
zachmullen: I'll open an issue for this and label it for later UI review.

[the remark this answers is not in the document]
jtomeck: Would something like this solve the associations?

proposed_search_grouping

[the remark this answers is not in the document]
zachmullen: @jtomeck pure genius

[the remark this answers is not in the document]
subdavis: @jtomeck oof
geicocaveman

src/components/Search.vue
subdavis: > is it easy to do this some other way

I originally did this another (messier) way, and @matthewma7 corrected it to this.  I agree with him that it's more readable this way. 

IMO that isn't an appropriate style for vuetify to apply for v-menu activators, and I might open a bug report about it.

src/components/Search.vue
```diff
@@ -0,0 +1,196 @@
...
+          v-icon {{ $vuetify.icons.alert }}
+        v-list-tile-content
+          v-list-tile-title No results found for query '{{ searchText }}'
+          v-list-tile-sub-title Modify search parameters or refine your query.
+      v-list-tile(v-show="!loading && showMore && searchResults.length > maxQuickResults",
+          @click="$emit('moreresults', searchParams)")
+        v-list-tile-action
+          v-icon {{ $vuetify.icons.more }}
+        v-list-tile-content
+          v-list-tile-title More
```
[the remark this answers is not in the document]
subdavis: It emits and event on line 98

subdavis:
@zachmullen I think I addressed your comments.  If you get a chance, could you make another pass and see if there's anything else?  Thanks!

src/components/Search.vue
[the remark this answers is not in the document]
zachmullen: Whoops!

src/components/Search.vue
zachmullen: From the Vuetify docs:

> open-on-click: Designates whether menu should open on activator click

IMO, if `open-on-click` is false, the cursor pointer shouldn't be used, so it's reasonable for us to file a bug report to see what they have to say. I do understand their rationale for this and I think it's a reasonable component for us to use, so I'm fine with overriding the style.

src/components/Search.vue
```diff
@@ -0,0 +1,192 @@
...
+  },
+};
+</script>
+
+<template lang="pug">
+v-layout.girder-searchbar(row, align-center)
+  v-icon.mdi-24px(color="white") {{ $vuetify.icons.search }}
+  v-menu.grow.mx-3(
+      offset-y, content-class="girder-searchbar-menu", :open-on-click="false",
+      :value="searchText", :nudge-bottom="6")
```
zachmullen: I tried out adding `transition="slide-y-transition"` and it looked pretty nice to me, should we add a transition to this menu opening? @jtomeck

src/components/Search.vue
```diff
@@ -0,0 +1,192 @@
...
+    @keyframes SkeletonShimmer {
+      0% { background-position: 0% 51%; }
+      50% { background-position: 100% 50%; }
+      100% { background-position: 0% 51%; }
+    }
+  }
+}
+
+// TODO: factor out into common stylesheet if we are to use this elsewhere.
+.girder-search-arrow-menu {
```
zachmullen: This whole block can be `scoped`, and I believe the `skeleton` stuff can be too.

src/components/Search.vue
```diff
@@ -0,0 +1,192 @@
...
+  transform: translateY(10px);
+  // Override to make the arrow visible
+  overflow: visible;
+  contain: inherit;
+
+  // Remove any shadow that is made visible by above style
+  :first-child {
+    box-shadow: none;
```
zachmullen: When I remove this rule I can't tell a difference, maybe it's just really subtle or I'm not on the right browser. What does this actually do?

src/utils/debouncer.js
```diff
@@ -0,0 +1,37 @@
...
+  inc() {
+    this.count += 1;
+    if (this.count > 0) {
+      this._toggleFlag(true);
+    }
+  }
+
+  dec() {
+    this.count -= 1;
+    if (this.count === 0) {
```
zachmullen: Should we clamp this to make sure it never goes below 0?

src/components/Search.vue
matthewma7: I added this originally. Vuetify v-menu has a shadow automatically and it prevent any kind of overflow to prevent a double shadow situation. The style to make the arrow visible to allow overflow causes the shadow of the v-card to show (technically really subtle to see).

src/components/Search.vue
matthewma7: Sounds good to me. I think I'll do it, @subdavis please pull before you work on other stuff.

src/utils/debouncer.js
subdavis: Sure, done.

src/components/Search.vue
subdavis: Looked fine to me, added it.  Can remove if @jtomeck feels otherwise.

subdavis:
@zachmullen could you approve again?  Had to rebase from master.

PR #82, opened 2019-02-14
Auth should forget password after successful login or session change.

subdavis:
Currently after login/logout, a user's password still exists in component state and can be retrieved with console access.

This adds 2 redundant behaviors:

* On successful login, the component will automatically clear the password field.
* On session change (such as a login) the entire component will be destroyed.   I think this is a good practice to also clear the username field and whatever other state has accumulated, but there may be reasons not to do this, so the decision is left up to the app developer.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -14,6 +14,7 @@ v-app.app
     girder-auth(
         :register="true",
         :oauth="true",
+        :key="girderRest.token",
         :forgot-password-url="forgotPasswordUrl")
   v-dialog(v-model="uploader", full-width, max-width="800px")
     girder-upload(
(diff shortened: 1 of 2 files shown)


src/App.vue
```diff
@@ -14,6 +14,7 @@ v-app.app
     girder-auth(
         :register="true",
         :oauth="true",
+        :key="girderRest.token",
```
zachmullen: What does this do? I've only seen `key` used with `v-for`

subdavis: It's just a generic way to control component lifecycle.  It's used with v-for as an efficient way of preserving elements through a re-render of the loop.

If key changes, the component is signaled to destroy and create a new instance of itself.   For sensitive components like login, seemed like the best way to clean up after itself.

zachmullen: Cool, good to know.

PR #84, opened 2019-02-14
Disallow '@' alias for local imports in src tree

zachmullen:
As promised, here's a linter rule that ensures we don't use the `@` alias within the `src` directory. Since the `vue-cli-service lint` command also lints the test directory, and we need that alias there, this rule cannot be applied to all directories. I assume it's also possible to put this in package.json and just override it in the test dir and any other dirs we need it, but this feels easier.

### DIFF
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -1,0 +1,5 @@
+{
+  "rules": {
+    "no-restricted-imports": ["error", { "patterns": ["@/*"] }]
+  }
+}


subdavis:
Follow up to #80

Awesome!  Thanks.

subdavis:
From the eslint docs:

> If there is an .eslintrc and a package.json file found in the same directory, .eslintrc will take a priority and package.json file will not be used.

I think this may be fully overriding the package json config.  We may need to put this into package json then override it in tests as you mentioned

zachmullen:
All set

PR #86, opened 2019-02-15
Separate demo

### LINKED ISSUE #77: Separate demo app from components library
At present, the demo app (bootstrapped by `src/main.js` and using components like `src/App.vue`) is located right next to the code for the components library (with its root at `src/index.js`).

Particularly considering that the demo app has hardcoded references to `data.kitware.com` and other deployment-specific details, it's good practice to strictly separate these two concerns. Although I (mostly) understand that the default rules for Vue-CLI will keep `src/main.js` out of the library if I import it downstream, having to rely on this undermines my confidence in this package as a deployment-agnostic library.

Can the demo app be moved to a separate directory, and can we ensure that the npm-published library does not include the demo app?

brianhelba:
* Gitignore built demo site files
* Move demo site files to a separate directory
* Set demo app API root from environment
* Ensure demo files are linted
  * Default "vue-cli-service lint" paths are in: https://github.com/vuejs/vue-cli/blob/v3.0.4/packages/@vue/cli-plugin-eslint/lint.js#L18

Fixes #77

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,9 +4,9 @@
   "license": "Apache-2.0",
   "main": "dist/girder.umd.min.js",
   "scripts": {
-    "serve": "vue-cli-service serve",
+    "serve": "vue-cli-service serve demo/main.js",
     "build": "vue-cli-service build --target lib --name girder src/index.js",
-    "build:demo": "vue-cli-service build --dest _site/",
+    "build:demo": "vue-cli-service build --dest _site/ demo/main.js",
     "lint": "vue-cli-service lint",
     "lint:pug": "pug-lint-vue src",
     "lint:style": "stylelint 'src/**/*.vue'",
--- a/package.json
+++ b/package.json
@@ -7,9 +7,9 @@
     "serve": "vue-cli-service serve demo/main.js",
     "build": "vue-cli-service build --target lib --name girder src/index.js",
     "build:demo": "vue-cli-service build --dest _site/ demo/main.js",
-    "lint": "vue-cli-service lint",
-    "lint:pug": "pug-lint-vue src",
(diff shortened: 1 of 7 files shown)


brianhelba:
Just pushed one more commit to ensure linting.

PTAL, this is ready for review.

package.json
```diff
@@ -4,12 +4,12 @@
...
   "scripts": {
-    "serve": "vue-cli-service serve",
+    "serve": "vue-cli-service serve demo/main.js",
     "build": "vue-cli-service build --target lib --name girder src/index.js",
-    "build:demo": "vue-cli-service build --dest _site/",
-    "lint": "vue-cli-service lint",
-    "lint:pug": "pug-lint-vue src",
-    "lint:style": "stylelint 'src/**/*.vue'",
+    "build:demo": "vue-cli-service build --dest _site/ demo/main.js",
+    "lint": "vue-cli-service lint src/ tests/ demo/ '*.js' '.*.js' '{src,tests,demo}/**/.*.js'",
```
zachmullen: Just tested locally and `"vue-cli-service lint src/ tests/ demo/"` seems to work, why are the other arguments necessary?

brianhelba: * `'*.js'` checks the root directory
  * for instance, `vue.config.js`
* `'.*.js'` and `'{src,tests,demo}/**/.*.js'"` checks hidden files, which ESLint does not pick up by default
  * for instance, `tests/unit/.eslintrc.js`

This logic comes from https://github.com/vuejs/vue-cli/blob/v3.0.4/packages/@vue/cli-plugin-eslint/lint.js#L18 , I just extended it to add `demo/`.

zachmullen: Seems reasonable.

PR #88, opened 2019-02-19
Presentation layer for Upload UI

subdavis:
Refactor some presentation-layer components for re-use in other projects.

This PR moves some of the UI from Upload.vue to their own stateless components.

I've tucked them into a subfolder called Presentation but I could potentially structure Upload like the Authentication folder.  I don't think that will be viable for much longer (and DataTable already provides an exception to the rule) but such things are easy to re-shuffle if we have a better idea later.

This PR benefits metabolomics because I can directly re-use those upload widgets.  This may invoke a larger discussion about what should and shouldn't be reused.  Since we don't have an org-wide library for purely presentation-layer components (and probably shouldn't at this point) I think this is pretty harmless.


src/components/Presentation/FileUploadList.vue
```diff
@@ -0,0 +1,56 @@
+<template lang="pug">
```
zachmullen: Why were the "Clear all" and "Start upload" buttons not moved into this component as well? Could you describe your use case and the rationale for breaking it up this way?

src/components/Presentation/Dropzone.vue
```diff
@@ -0,0 +1,94 @@
+<template lang="pug">
+  .dropzone-wrapper(
+      v-if="!files.length", :class="dropzoneClass", @dragenter="dropzoneClass = 'animate'",
+      @dragleave="dropzoneClass = null", @drop="dropzoneClass = null")
+    v-layout.dropzone-message(column, justify-center, align-center, fill-height)
+      v-icon(size="50px") $vuetify.icons.fileUpload
+      .title.mt-3 {{ message }}
+    input.file-input(type="file", :multiple="multiple", @change="$emit('change', $event)")
```
zachmullen: Should this component have `v-model` support for the files? Seems like it's always going to be two-way binding from some parent component.

subdavis: Yes, I agree. I'll add this.

src/components/Presentation/FileUploadList.vue
subdavis: I broke this out separately because I think smaller and more specialized components are more flexible.  The composition of these components implies a particular intent "I want to upload 1 or more files to a server, and I don't care what happens to the files when they're done uploading".

A dropzone is useful to me outside the context of an upload widget because I could read the files into client memory and immediately produce a visualization of the data. 

I'm actually using this in the metabolomics UI with dropzone and fileuploadlist, but without the clear and upload buttons because those controls are elsewhere on the page.  The upload widget is part of a larger form, and having those buttons in the widget as they are here is like putting the submit button in the middle of a form: I want to trigger the upload and form submit together as a single operation.   I want to hide the idea of "uploading files" from the user and present a feel of "processing data" without having to care that a server is involved.

Finally, those buttons are very low-effort to reproduce in another app.  I thought including them would reduce the flexibility to reuse the file list and add almost no value re: preventing code duplication.

src/components/Presentation/Dropzone.vue
subdavis: Done.  Added an immutable splice function.

zachmullen: I see you added it to the FileUploadList, but I was wondering if it makes sense to add it to the Dropzone component as well? (Since it's an input)

subdavis: Oops, yeah that does make sense.

subdavis:
I actually don't think `v-model` makes sense for dropzone.  We don't use `v-model` on the underlying file input box, it just emits change events.  I removed `files` as a prop to dropzone since it was only being used for a `v-if`

subdavis:
@zachmullen required rebase, could you reapprove?  ty.

Issue #89, opened 2019-02-21
Pressing "enter" when focused in folder name should submit form


zachmullen:
In the folder creation dialog, typing a name and hitting "enter" should send the XHR to create the folder. As it stands, the "Create Folder" button must be clicked.

PR #91, opened 2019-02-21
Use submit event and enable browser autocomplete

matthewma7:
Would resolve https://github.com/girder/girder_web_components/issues/89

I personally like browser autocomplete so I enabled autocomplete on the folder name. If we like it we could potentially enable on all common input fields, if not I can remove the autocomplete.

### DIFF
--- a/src/components/UpsertFolder.vue
+++ b/src/components/UpsertFolder.vue
@@ -106,36 +106,38 @@ export default {
 </script>
 
 <template lang="pug">
-v-card(flat, height="100%")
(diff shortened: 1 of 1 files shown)


zachmullen:
I've verified that this fixes #89. 👍 

I'm not a fan of the autocomplete behavior for this case personally.

matthewma7:
Hm, I guess folder name is not complex enough to be auto completed. If it is an URL, it probably will worth it. 
I removed the autocomplete.

PR #92, opened 2019-02-22
pull merge up to version 1.2.1

subdavis:
Pull merge up to version 1.2.1.


zachmullen:
What's this for?

subdavis:
screenshot_20190222_103232

Don't totally understand the threat (and I think the risk is using it server-side) but I figure it's easier to upgrade than figure out exactly what's going on.

zachmullen:
Yeah, the vulnerability doesn't affect us but this is definitely the Right Thing To Do™.

Issue #65, opened 2019-01-22
File Details Component


subdavis:
In #31 @jtomeck described a file details view.  The renders are in the zip file mentioned there.

Here's a rough example for how that could look.  I haven't put a ton of work into this yet, and the action handlers aren't wired up to anything.   If we were to add this, we would also eventually need a bulk file-move dialog and a bulk copy dialog.

Rather than the pick/drop behavior in the current girder client, I would probably lean toward something like the bulk-move-copy feature on dropbox.com.  It seems a little friendlier.

#### Single file operations:

![screenshot_20190122_162507](https://user-images.githubusercontent.com/4214172/51567179-9660cd00-1e64-11e9-92a9-27e55e7f764e.png)

#### Bulk operations:

![screenshot_20190122_162552](https://user-images.githubusercontent.com/4214172/51567180-9660cd00-1e64-11e9-8bea-506fa3135028.png)

As an aside, I've noticed that a lot of our look-and-feel is similar to the file browser on DropBox.  Perhaps it would be good to discuss the behavior of some of these components in terms of how DropBox/Box/GDrive/OwnCloud behave and find other examples of widgets we can lift from those interfaces.

PR #93, opened 2019-02-25
Details View

### LINKED ISSUE #163: Add in detail view to hierarchy widget?
Does it make sense to have a detail view in the hierarchy widget like in the attached screenshot? Can we make it so that the actions are extensible?

Should this be a feature in the general library or be project specific?

Screenshot from 2019-10-02 14-59-17

subdavis:
closes #163 #65 

I've been sitting on this branch a while, recently rebased it since everything else got merged.

It isn't ready to go in, but it is neat to play with as we talk about what else this library should have.  I think the introduction of this component exposes some of the weak points in the current state management since state flows between browser and details, and both can modify state.

After this component, I don't know what the next feature for this library will be.  It might be good to let it breathe and exercise it more as consumers from other apps to get a feel for what we've done well and what can be improved.

Feel free to browse through or totally ignore this PR.  I'm only opening it for the sake of sharing and discussion.

### DIFF
--- a/src/components/DataDetails.vue
+++ b/src/components/DataDetails.vue
@@ -19,8 +19,8 @@ v-card.data-details
   girder-detail-list(v-if="meta.length", title="Meta", :rows="meta")
     template(#row="props")
       v-row(justify="space-between")
-        v-col.shrink.py-1 {{ props.datum.key }}
-        v-col.shrink.py-1 {{ props.datum.value }}
+        v-col.shrink.py-1.body-2.font-weight-bold {{ props.datum.key }}
+        v-col.shrink.py-1.body-2 {{ props.datum.value }}
   girder-detail-list(
       v-if="files.length",
       :title="`Files (${files.length})`",
@@ -28,16 +28,14 @@ v-card.data-details
   girder-detail-list(
       v-if="actions.length",
       title="Actions",
+      :clickable="true",
       :rows="actions",
-      :color="$vuetify.theme.themes.light.secondary.lighten5")
+      :color="$vuetify.theme.themes.light.secondary.lighten5",
+      @click="handleAction")
     template(#row="props")
-      v-row
-        v-col.px-2.py-0
-          v-btn.pl-1(text, small, block,
-              :color="props.datum.color",
-              @click="handleAction(props.datum)")
-            v-icon.pr-2 {{ $vuetify.icons.values[props.datum.icon] }}
-            | {{ props.datum.name }}
+      v-list-item-icon.mr-1
+        v-icon.pr-2(:color="props.datum.color") {{ $vuetify.icons.values[props.datum.icon] }}
+      v-list-item-content(:class="`${props.datum.color}--text`") {{ props.datum.name }}
 </template>
 
 <script>
@@ -254,9 +252,3 @@ export default {
   },
 };
 </script>
-
-<style lang="scss">
-.data-details .v-toolbar__content {
-  padding: 0 16px !important;
-}
-</style>
--- a/src/components/DataDetails.vue
+++ b/src/components/DataDetails.vue
@@ -1,0 +1,262 @@
+<template lang="pug">
+v-card.data-details
+  v-toolbar(flat, dark, dense, color="primary")
+    v-toolbar-title.subtitle-1
+      v-icon.pr-2.mdi-18px {{ icon }}
+      | {{ title }}
+    v-spacer
+    v-dialog(v-if="datum", v-model="showUpsert", max-width="800px")
+      template(v-slot:activator="{ on }")
+        v-btn(v-on="on", icon)
+          v-icon.mdi-18px {{ $vuetify.icons.edit }}
+      girder-upsert-folder(
+          :edit="true",
+          :key="datum._id",
+          :location="datum",
+          @dismiss="showUpsert = false")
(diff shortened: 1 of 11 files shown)


subdavis:
Screenshots: 

Screenshot from 2019-10-02 14-57-28
Screenshot from 2019-10-02 14-57-40

subdavis:
@zachmullen Upgraded to vuetify 2.0.  

It isn't perfect, but you should be able to use it while I think about it some more.  Extend the actions and info list by importing the constant arrays and adding/removing entries.    These are intended to work kinda like the `headers` prop of `v-data-table`.  Let me know if you have questions.

zachmullen:
Thanks, is there some kind of pre-release I can install off of this branch? I guess I can install from a local package, but I forgot the incantation to do so

subdavis:
Nope, we don't have any kind of build strategy to publish pre-releases from PRs (maybe we should?)

You can however change the package json dependency string to `git@github.com:girder/girder_web_components#feature/details`

zachmullen:
For posterity, the exact package.json string is `"@girder/components": "girder/girder_web_components#feature/details"`

src/components/Snippet/FileManager.vue
```diff
@@ -148,62 +151,59 @@ export default {
 </script>
 
 <template lang="pug">
-.girder-data-browser-snippet
-  v-row
-    v-col
-      v-card
```
subdavis: @matthewma7 what's the reason for the `v-row`, `v-col`, `v-card` nesting here?  I removed them and everything looks the same.

matthewma7: Hm, I don't remember, but please go ahead and remove it if it works fine without them.

subdavis:
@zachmullen could you make a quick pass over `DataDetails.vue` for any glaring code quality issues?  If you already have, I'll just add a couple tests and get this merged.

src/components/DataDetails.vue
```diff
@@ -0,0 +1,274 @@
...
+</script>
+
+<style lang="scss" scoped>
+.data-details {
+  width: 100%;
+
+  .details-title {
+    font-size: 16px;
+  }
+}
```
zachmullen: Why is `width: 100%` necessary?

Also, could we use one of the semantic classes from vuetify rather than hard-coding a font-size?

subdavis: Done.  Also added a kinda trivial test as little more than a sanity check. I'll add some more later.

zachmullen: Going from nothing to a "does it explode or not" test has a great ROI. Good stuff.

src/components/Presentation/DetailList.vue
```diff
@@ -0,0 +1,62 @@
...
+
+  .no-overflow {
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow: hidden;
+    width: 100%;
+  }
+
+  .v-btn__content {
+    justify-content: flex-start;
```
zachmullen: Given that there are no `v-btn`s in this, this style rule doesn't belong here. I'm actually working on a proposed change that will remove all these rules, I'll push that up to this branch for you to review.

subdavis: Awesome, I hadn't really reviewed these styles since I wrote them back in February.  Looks like most of them are reproducible with Vuetify conventions.  I'll wait to make any changes until I've seen yours.

zachmullen: 4fe54a9 is my suggested change, it has the benefit of removing all the CSS from the reusable components.

I think there's an argument to be made that the entire `v-list-item` inside DataDetails should be inside of the slot rather than just its contents, to allow for things like the `to` attribute, or any other attributes of the `v-list-item` that someone might want to control from outside. What do you think?

subdavis: To be honest, I'm not sure `v-list` is appropriate at all here.   We aren't really using any of its features except for the buttony thing. `v-list` produces lists that generally feel more "important" than simple lists of text (navigation bars and the like).

I think we should keep your changes for now though, until some more advanced UI need presents itself.

src/components/Presentation/DetailList.vue
```diff
@@ -0,0 +1,33 @@
+<template lang="pug">
+v-list(v-if="rows.length", dense, :style="{ backgroundColor: color }")
+  v-subheader.body-2.font-weight-bold {{ title }}
```
subdavis:
```suggestion
  v-subheader.subtitle-1.font-weight-bold.pl-4 {{ title }}
```

I would like to bump the size on those subheaders and vertically align them with the content, if that's alright.

zachmullen:
One other feature we will need on the DataDetails component is the ability to also extend the file-specific metadata section, but I'd like to defer that to another PR.

subdavis:
Will you need to make buttons and stuff based on metadata?  What sort of extension?

zachmullen:
My specific use case is that I need to show an additional metadata field from the file, it's just plaintext, might have a transform function like the item metadata.

subdavis:
Did the girder plugin add a meta field to files?  I wasn't aware that they had metadata.

zachmullen:
By metadata I mean system-defined metadata (just a top-level field of the file document)

subdavis:
ed16474 should be the last change needed here.

PR #94, opened 2019-03-05
Allow Upload.vue to take optional upload override

subdavis:
This PR is motivated by @waxlamp so that he can override the default upload behavior and use the whole component potentially to target a different upload server (potentially with some other non-standard behavior).

After we spoke, I don't see any reason why something like this wouldn't work, but I'm not sure if we're solving his problem in the right place.

### DIFF
--- a/src/components/Upload.vue
+++ b/src/components/Upload.vue
@@ -60,6 +60,10 @@ export default {
       default: () => {},
       type: Function,
     },
+    upload: {
+      default: undefined,
+      type: Function,
+    },
   },
   data: () => ({
     dragover: false,
@@ -105,13 +109,15 @@ export default {
       }));
     },
 
-    async start() {
+    /**
+     * @param files the files to upload
+     * @returns {results, err: {error, file, message}}
+     */
+    async _upload(files) {
       const results = [];
-      this.uploading = true;
-      this.errorMessage = null;
-      await this.preUpload();
-      for (let i = 0; i < this.files.length; i += 1) {
-        const file = this.files[i];
+
+      for (let i = 0; i < files.length; i += 1) {
+        const file = files[i];
         if (file.status === 'done') {
           // We are resuming, skip already completed files
           results.push(file.result);
@@ -133,22 +139,40 @@ export default {
             results.push(file.result);
             file.status = 'done';
           } catch (error) {
+            let message = '';
             if (error.response) {
-              this.errorMessage = error.response.data.message || 'An error occurred during upload.';
+              message = error.response.data.message || 'An error occurred during upload.';
             } else {
-              this.errorMessage = 'Connection failed.';
+              message = 'Connection failed.';
             }
             file.status = 'error';
-            this.uploading = false;
-            this.$emit('error', { error, file });
-            return;
+            return { results: null, err: { error, file, message } };
           }
         }
       }
-      await this.postUpload();
+      return { results, err: null };
+    },
+
+    async start() {
+      this.uploading = true;
+      this.errorMessage = null;
+
+      await this.preUpload();
+
+      const { results, err } = this.upload !== undefined
+        ? await this.upload(this.files)
+        : await this._upload(this.files);
+
       this.uploading = false;
-      this.files = [];
-      this.$emit('done', results);
+
+      if (err !== null) {
+        this.errorMessage = err.message;
+        this.$emit('error', err);
+      } else {
+        await this.postUpload();
+        this.files = [];
+        this.$emit('done', results);
+      }
     },
   },
 };


subdavis:
Marked draft because I want to add more tests if this is OK.

zachmullen:
When I had considered this last week, I had thought of hooking into a different place. I don't know which is better, but what I had considered was allowing the `Upload` class to be a prop:

```javascript
import Upload from '../utils/upload';
...

props: {
  uploadClass: {
    type: Function,
    default: Upload,
  }
}
```

That isn't as powerful as what you've written, but is more friendly if are fine with the same sequential upload behavior for multiple files; downstreams would only have to implement a Promise-based Upload class that handled a single file, and wouldn't need to worry about overall progress, etc.

One issue I ran into with my approach is the API of `Upload`; it expects a `RestClient` as the first argument, and we wouldn't want that API for non-Girder uses (a single options Object would be preferable). We are free to make that change since we're still in alpha.

@waxlamp what behaviors do you need exactly?

waxlamp:
I was tinkering with the idea of using the upload component to perform a completely non-Girder upload: to push data of a specific format into, e.g., a NoSQL database. (This was also something of a thought experiment following on from our Friday discussion about separating these components into presentation and container layers.)

subdavis:
> That isn't as powerful as what you've written

I don't really see a difference in the level of control these allow (aside from sequential upload).  I guess downstream couldn't programatically resume if it detected a change in network conditions with the class-based solution.

> wouldn't need to worry about overall progress

It doesn't seem like either required downstream to care about overall progress.

> if users are fine with the same sequential upload behavior for multiple files

That sounds reasonable to me.

jeffbaumes:
Not knowing much of the nuance here, if we are choosing between a callback function prop and a "callback" class prop that needs to implement a particular API, a callback function prop sounds simpler to use and more in line with what Vue devs would expect. Are there other other known cases of passing classes to Vue components as props?

zachmullen:
> Are there other other known cases of passing classes to Vue components as props?

I think using props as dependency injection is reasonable, but can't speak to how familiar it is. The concept of service dependency injection in general should be very familiar to Vue devs, due to things like `inject`/`provide` and the way its plugin system works.

subdavis:
@jeffbaumes or @zachmullen do either of you have a strong preference?  I see merits in both.

* Class approach seems like a nice separation of concerns.  Dependency injection through props is a little irregular, but I don't see anything wrong with it.
* simple callback feels more generic, and could potentially cover more use cases.

I'm leaning toward to the simple callback approach, but only slightly.

waxlamp:
I don't fully understand the benefits of the "class injection" approach, but I largely agree with Jeff: I like the simplicity and familiarity of using a function.

zachmullen:
I'd like to put together a code snippet to show how extension of the `Upload` class could make this quite straightforward. It could also handle the `preUpload` and `postUpload` hooks, in addition to adding hooks before and after each file upload.

zachmullen:
That said, I don't want to hold this up. We can merge this if it meets the immediate need, and consider adding another hook prop later.

brianhelba:
> we wouldn't want that API for non-Girder uses

> I was tinkering with the idea of using the upload component to perform a completely non-Girder upload

So is `girder_web_components` now supporting components that are supposed to be generalized enough to interact with things other than Girder?

waxlamp:
I think it's an important thing to consider, especially if the design supports the 100% Girder use case out of the box, and can adapt to unforeseen use cases easily and effectively.

My use case has us using Girder as the API framework because Girder "does user management" already. If the use case requires uploading to something that isn't a Girder assetstore, but we still want to mediate access control through Girder's user system (i.e. my discussion from yesterday), then having GWC components that are flexible enough to work in this more general realm is something that lowers the activation energy of using Girder in the first place.

brianhelba:
GWC provides an `Authentication` component(s) for user management and a `RestClient` to interact with Girder in an even more general way.

If a widget is needed to upload files to somewhere that's not Girder, why wouldn't a new, downstream widget be called for? It could even internally use `RestClient` to the extent necessary.

It'd be instructive to me for @subdavis @zachmullen @jeffbaumes to provide some upstream guidance on what the scope of GWC is intended to be. What functionality isn't on the table to be further generalized or have explicit hooks added for overriding? I don't blame @waxlamp in the slightest for asking for features (and maybe this one is indeed worthwhile), but will upstream say no to certain things?

I care a lot more about this general question than the particulars of the PR, so apologies if we need to move this to an issue to prevent the PR from getting hijacked.

zachmullen:
> but will upstream say no to certain things?

That's an easy one to answer: yes, we would say no to certain things.

If you're asking for a comprehensive, concise policy about things we would say no to, that's not something I could provide. Things will be evaluated on a case by case basis. The best I could offer is a methodology for how we decide what things to say yes or no to, which is something like a CBA. That's precisely the code review process, isn't it? Answering the question: "Is this piece of code more valuable than it is costly?" Maybe this is obvious, but I'm sort of having to guess at the intent of the line of questioning.

To answer quite directly, yes, the core stakeholders of this library reserve the right to say no to changes if we think they would introduce net negative value to the codebase.

zachmullen:
To more affirmatively answer your question and attempt to define the scope of this library, I would say this about this PR: adding this hook to support @waxlamp 's current use case should *not* be construed as us supporting non-Girder usages of our components in general. This change comes with no implicit promise about future support of non-Girder uses of this or any other components in the library. It should be interpreted very narrowly if it passes QA: we feel that a hook to control the upload behavior is within scope; the fact that that helps @waxlamp use this in a non-Girder way is the motivator of this change, but the change itself is not causing us to expand the scope of GWC in any meaningful way. I personally have existing (Girder-backed) use cases where I will need to hook into a few points during the upload process to add custom behavior, and this is one mechanism that could address that, another being my proposed Upload class prop.

I would add that if you personally have objections to this code change in and of itself, just in the context of GWC's scope as you understand it, please do voice them during the review. Even if you are just a user rather than a maintainer, your opinion is still valuable.

waxlamp:
On Wed, Mar 6, 2019 at 2:47 PM Zach Mullen <notifications@github.com> wrote:

> That said, I don't want to hold this up. We can merge this if it meets the
> immediate need, and consider adding another hook prop later.
>
I'm ok to "hold up" until you can put together a snippet.

matthewma7:
I have a use case to use a custom upload (Don't upload a girder item but call a custom girder endpoint to create custom girder items). So, seems this use case is prevalent.

When I read the implementation of the _upload(), I noticed it is doing a lot of things I didn't aware it was doing. Such as the stuff like [progress](https://github.com/girder/girder_web_components/blob/f3fdc97301d7e289ca1deb7ac7310768abfd3a99/src/components/Upload.vue#L134)? At this moment, without really reading the code, I am not confident to provide an overriding update()

Now I totally understand what @zachmullen means by

> but is more friendly if users are fine with the same sequential upload behavior for multiple files; downstreams would only have to implement a Promise-based Upload class that handled a single file, and wouldn't need to worry about overall progress

I personally do expect `only have to implement a Promise-based Upload class that handled a single file`. 

So, @zachmullen, is a class a must in your case, would a pair of start() and resume() or even just a start()  work? 

@zachmullen and @all, about the progress, can we provide the start() with a callback as an argument, which would expect a simple 0 to 100 number, but if not called. Just assume the file is finished when start() returns.

subdavis:
Closing in favor of #99

PR #99, opened 2019-03-15
Upload subclassing

zachmullen:
This isn't intended to be merged as is, just to show a rough idea of what I'm talking about on #94.

With this interface, downstreams could subclass the Upload class and pass it into the Upload component to control behavior. This also adds 3 hook points into the upload class that are no-ops by default. These hooks would allow things like:

* Perform an action (synchronous or asynchronous) before the first upload.
* Perform an action before *each* upload, or only certain uploads, based on filename or other file properties of the file.
* Perform an action after the *last* upload is done.
* Perform an action if an error occurs during upload.

If we allow these hooks to return values, they could also be used to further augment the behavior of the component, e.g. skipping certain files, or showing custom errors. This concept would need to be fleshed out further, but we don't have to do it right now, as we could introduce that kind of control without breaking this API.

Note that the need for `preUpload` and `postUpload` callback props on the component is obviated.

If the general idea is palatable, I'll write a unit test that creates a subclass and exercises the hooks.

### DIFF
--- a/src/utils/upload.js
+++ b/src/utils/upload.js
@@ -90,4 +90,7 @@ export default class Upload {
 
   // eslint-disable-next-line class-methods-use-this
   afterUpload() {}
+
+  // eslint-disable-next-line class-methods-use-this
+  onError() {}
 }
--- a/src/utils/upload.js
+++ b/src/utils/upload.js
@@ -88,10 +88,7 @@ export default class Upload {
   /**
    * This callback is called before the upload is started. This callback is asynchronous.
    * If it returns a Promise, the caller will await its resolution before continuing.
-   * This callback is passed the index of this file in the overall list of files being uploaded,
-   * which is mostly done to allow asynchronous hooks to be performed only before the first file or
-   * after the last file, which is a common use case.
-   * @param current {Number} The index of this file in the list.
+   * @param current {Number} The index of this file in the list of files being uploaded.
    * @param total {Number} The total number of files being uploaded.
    */
   // eslint-disable-next-line class-methods-use-this, no-unused-vars
@@ -100,10 +97,7 @@ export default class Upload {
   /**
    * This callback is called after the upload is completed. This callback is asynchronous.
    * If it returns a Promise, the caller will await its resolution before continuing.
-   * This callback is passed the index of this file in the overall list of files being uploaded,
-   * which is mostly done to allow asynchronous hooks to be performed only before the first file or
-   * after the last file, which is a common use case.
-   * @param current {Number} The index of this file in the list.
+   * @param current {Number} The index of this file in the list of files being uploaded.
    * @param total {Number} The total number of files being uploaded.
    */
   // eslint-disable-next-line class-methods-use-this, no-unused-vars
@@ -112,9 +106,8 @@ export default class Upload {
   /**
    * This callback is called if an error occurs during the upload. This callback is asynchronous.
    * If it returns a Promise, the caller will await its resolution before continuing.
-   * This callback is passed the index of this file in the overall list of files being uploaded.
    * @param error {Exception} The exception object.
-   * @param current {Number} The index of this file in the list.
+   * @param current {Number} The index of this file in the list of files being uploaded.
    * @param total {Number} The total number of files being uploaded.
    */
   // eslint-disable-next-line class-methods-use-this, no-unused-vars
--- a/src/utils/upload.js
+++ b/src/utils/upload.js
@@ -7,16 +7,18 @@ const uploadBehaviors = { s3: S3Upload };
 export default class Upload {
   /**
    * Represents an upload of a single file to the server.
-   * @param $rest {Object} an axios instance used for communicating with Girder.
    * @param file {File | Blob} the file to upload
-   * @param parent {Object} upload destination. Must have ``_id`` and ``_modelType`` properties.
    * @param opts {Object} upload options.
+   * @param opts.$rest {Object} an axios instance used for communicating with Girder.
+   * @param opts.parent {Object} upload destination. Must have ``_id`` and ``_modelType``.
    * @param opts.progress {Function} A progress callback for the upload. It will receive an Object
-   * argument with either ``"indeterminate": true``, or numeric ``current`` and ``total`` fields.
+   *   argument with either ``"indeterminate": true``, or numeric ``current`` and ``total`` fields.
    * @param opts.params {Object} Additional parameters to pass on the upload init request.
    * @param opts.chunkLen {Number} Chunk size for sending the file (integer number of bytes).
    */
-  constructor($rest, file, parent, {
+  constructor(file, {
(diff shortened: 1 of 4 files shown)


matthewma7:
Thank you @zachmullen for providing this concrete example. I don't know if this is too much to ask, but I think if it's possible to pass the [parameters](https://github.com/girder/girder_web_components/blob/77082e804fe837fded1fe86681f343174c6c8e7d/src/utils/upload.js#L19-L23) to each handler (start, resume, onerror, etc) so no class is needed, and could handle the situation that the implementation doesn't provide progress (files too small to matter) it would be great. 
Bascially making things optional, not required, a class, resume(), progress().

zachmullen:
Progress is already optional, you don't have to call the progress function if you don't want.

I'm curious, why would you prefer duplicating and passing the entire state on each event rather than allowing the object to store what state it needs? To me the class pattern is much more elegant and intuitive.

matthewma7:
I think I personally feel a class feels a little bit foreign in JavaScript, unlike in language like Java, passing a function (delegate or whatever it's called) would feel foreign, but this also has something to do with the framework, for things like OpenLayers and stuff, an overriding class is more common than an overriding function, no any technical reason. I think the more important part of this feature is making things optional. (only provide start(), not even no-op handlers)

I tried to comment out everything related to `this.progress` callback in the built-in upload.js, the progressbar will not progress and the `v-progress-circular` won't move, so I assumed maybe something else is needed to handle that.

zachmullen:
> I think I personally feel a class feels a little bit foreign in JavaScript

Maybe so, I may be in the minority here as someone who came up during the OO craze. I'm not married to this idea -- feel free to write up your own draft PR. The reason I originally chose a class to handle this is because the upload is pretty stateful, and to downstreams, the fact that it's a class was hidden -- using it looked identical to old-style constructors that have been around JS forever.

> I tried to comment out everything related to this.progress callback in the built-in upload.js, the progressbar will not progress and the v-progress-circular won't move, so I assumed maybe something else is needed to handle that.

As part of this, we would just need to have the Upload.vue component provide sensible defaults, i.e. indeterminate at the start, and 100% on completion. Right now we rely on the upload class to do that, and I didn't make that change on this PR yet.

matthewma7:
Totally, I didn't mean to ask you to actually implement it. 
I have a need for this feature at this moment. I'll start with this branch and modify to have a draft. Should be too hard with what we have here. 🙂

waxlamp:
This looks pretty cool to me.

As someone who also grew up during the OO era, but has recently been intrigued by the functional programming crazy, I generally prefer functions as well, but I don't feel like using a class is a problem as in this strawman example. Overall, I don't have strong feelings either way.

I think the important part is to allow this to
1. replicate current behavior through sensible defaults, etc.
2. be extensible in the future without breaking the expanding API we're introducing here

In particular, I like the idea of allowing this to behave "monadically" in the future (that is, determine future behavior based on return values from intermediate actions) but it would be something to consider carefully, if we decide to add that at all, and it would need to be driven by actual needs (whether from customers or ourselves).

Does this design support overriding the `start()` method with a no-op, in case you want to get weird and not do the actual Girder upload at all (as in my initial use case for this stuff)?

I do feel like this approach allows us both extremes: to mold file upload into something alien that we wouldn't have thought anyone would want to do, while leaving the common case intact, and providing a few hooks to serve cases that fall in between (while leaving ourselves a path to provide *more* such hooks as demands arise). If that way of thinking about this misses anything important, it would be good to bring that to light.

Thanks for putting this together, Zach.

zachmullen:
> Does this design support overriding the start() method with a no-op, in case you want to get weird and not do the actual Girder upload at all (as in my initial use case for this stuff)?

Yes, the return value of `start()` can be anything you want, the Upload component itself does nothing with it other than trigger an event containing the values after all uploads are complete. So you can just have:

```javascript
class MyUpload extends Upload {
    start() {}
}
```

matthewma7:
This upload subclassing worked great for my case.
I believe my subsequential PR https://github.com/girder/girder_web_components/pull/102 make this change backward compatible and more easy to use. 

I would like to push this feature forward. 
@zachmullen @subdavis What do you think?

@zachmullen I can fix the tests here if you would like.

src/utils/upload.js
```diff
@@ -82,4 +84,13 @@ export default class Upload {
     this.offset = (await this.$rest.get(`file/offset?uploadId=${this.upload._id}`)).data.offset;
     return this._sendChunks();
   }
+
+  // eslint-disable-next-line class-methods-use-this
+  beforeUpload() {}
```
subdavis: Might be good to document the arguments this takes.

src/utils/upload.js
```diff
@@ -7,16 +7,18 @@ const uploadBehaviors = { s3: S3Upload };
 export default class Upload {
   /**
    * Represents an upload of a single file to the server.
-   * @param $rest {Object} an axios instance used for communicating with Girder.
    * @param file {File | Blob} the file to upload
-   * @param parent {Object} upload destination. Must have ``_id`` and ``_modelType`` properties.
    * @param opts {Object} upload options.
+   * @param opts.$rest {Object} an axios instance used for communicating with Girder.
```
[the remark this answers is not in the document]
zachmullen: Based on @waxlamp 's use case, some downstreams may not even pass a RestClient, so we made all params optional.

zachmullen:
This is ready for review.

src/utils/upload.js
```diff
@@ -82,4 +84,39 @@ export default class Upload {
...
+   * @param current {Number} The index of this file in the list.
+   * @param total {Number} The total number of files being uploaded.
+   */
+  // eslint-disable-next-line class-methods-use-this, no-unused-vars
+  afterUpload({ current, total }) {}
+
+  /**
+   * This callback is called if an error occurs during the upload. This callback is asynchronous.
+   * If it returns a Promise, the caller will await its resolution before continuing.
+   * This callback is passed the index of this file in the overall list of files being uploaded.
```
subdavis: Move any relevant info from this line into the @param line and remove?

src/utils/upload.js
```diff
@@ -82,4 +84,39 @@ export default class Upload {
...
+   */
+  // eslint-disable-next-line class-methods-use-this, no-unused-vars
+  beforeUpload({ current, total }) {}
+
+  /**
+   * This callback is called after the upload is completed. This callback is asynchronous.
+   * If it returns a Promise, the caller will await its resolution before continuing.
+   * This callback is passed the index of this file in the overall list of files being uploaded,
+   * which is mostly done to allow asynchronous hooks to be performed only before the first file or
+   * after the last file, which is a common use case.
```
subdavis: Move any relevant info from 103 line into the @param line and remove?

Remove 104, 105, and also from 91,92.  I think it's fairly clear how to use these hooks.  

Or maybe put it into the header docstring for this file?

PR #102, opened 2019-03-19
Upload class optional members

matthewma7:
This PR contains some maintenance changes, make all callback optional except the start() method, and make sure progress updates when the callback is not called. 

This PR also add back removed callbacks because the callback was applied before and after all files and can be easily used on consuming component (instead of inside uploader).

### DIFF
--- a/src/components/Upload.vue
+++ b/src/components/Upload.vue
@@ -119,10 +119,11 @@ export default {
             Object.assign(file.progress, event);
           };
           file.status = 'uploading';
+          file.progress.indeterminate = true;
           try {
             if (file.upload) {
               // eslint-disable-next-line no-await-in-loop
-              file.result = await file.upload.resume();
+              file.result = await (file.upload.resume ? file.upload.resume() : file.upload.start());
             } else {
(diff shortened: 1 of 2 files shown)


subdavis:
I think it keeps things simple to make the callbacks required.  Fewer conditionals and some unused method stubs to me is preferable.

I can certainly see that the overall before/after callbacks would be useful, but I don't know that I want to maintain both patterns.  As a user, I've already gone to the trouble of setting up my own Upload class with access to the `$rest` object I'll likely need in my before/after hooks.  I'd rather not have to deal with accessing state from both these component hooks and my class hooks.

I think @zachmullen already provided an alternative by reading `current` and `total` from inside the `beforeUpload` and `afterUpload` handlers.   Could you use those instead?

subdavis:
#99 was merged

zachmullen:
Ok to abandon this, 👎 on the conditionals.

matthewma7:
Wait, sorry if I missed my chance.
But I think this PR provided something else. 
I'll create new PR to talk about my point.

PR #106, opened 2019-04-03
Adds draggability flag

subdavis:
This feature was requested by @zackgalbreath and was pretty trivial to implement.

* Add dragability for rows in DataTable.vue
* Upstream is expected to register and use these rows with some dropzone.
* New events:  DataBrowser and DataTable both emit:
  * drag
  * drop
  * dragstart
  * dragend
* New props:
  * DataBrowser gets `draggable: Boolean`
  * DataTable gets `draggable: Boolean`

Those event payloads are structured `{ items: Array<{ index, item, selected }>, event: <native event> }` .

@zackgalbreath will this suit your use case?  I actually really like this as a feature and I may be able to use it myself soon.

### DIFF
--- a/demo/App.vue
+++ b/demo/App.vue
@@ -15,6 +15,7 @@ v-app.app
           v-card-actions
             v-layout(column)
               v-checkbox.mt-2(hide-details, label="Select", v-model="selectEnabled")
+              v-checkbox.mt-1(hide-details, label="Draggable", v-model="dragEnabled")
               v-checkbox.mt-1(hide-details, label="New Folder", v-model="newFolderEnabled")
               v-checkbox.mt-1(hide-details, label="Upload", v-model="newItemEnabled")
               v-checkbox.mt-1.mb-1(hide-details, label="Search Box", v-model="searchEnabled")
@@ -50,6 +51,7 @@ v-app.app
               v-if="location",
               :location.sync="location",
               :select-enabled="selectEnabled",
+              :draggable="dragEnabled",
               :new-item-enabled="newItemEnabled",
               :new-folder-enabled="newFolderEnabled",
               @click:newitem="uploader = true",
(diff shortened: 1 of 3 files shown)


subdavis:
> dragging one of them should drag all the checked rows.

I think that would be a great feature to have, but it isn't natively supported as far as I'm aware.  HTML5 `draggable` will only let you pick up a single element.

* [Native example with 1 thing](https://www.w3schools.com/html/tryit.asp?filename=tryhtml5_draganddrop)
* JS hack example with multiple things: try dragging multiple selections in Gmail.  In my browser, it looks like google is listening to mouse move events and moving an absolute element around on the page.  The cursor hand also flickers because I think native styling and JS are fighting.

We could certainly add our own implementation, but there are already libraries like https://shopify.github.io/draggable/ I believe it would require significantly more work to integrate one of those libraries than the 20 or so minutes I spent on this PR

Another option is just to let that additional selection data flow through the event, so it looks like you're only dragging 1 thing, but the event contains the data for all selected rows.  I'm really not a fan of this because 1) I think it will confuse the user and 2) any app could already do this without a change to the API because `Array<selected_items>` is already exposed by the DataBrowser:

```javascript
if (dropped item was selected) {
  do stuff with databrowser.selected array
} else {
  just do stuff with item
}
```

Because Zach G. Doesn't need to drag multiple things for his application, I think I'd like to stick with the native approach for now and revisit if new use cases emerge.

----------

EDIT:  Also, using native draggability makes the feature much more accessible to consumers because they need only configure a dropzone.  If we have to do some client implementation, we either have to make upstream learn/use the same library we did (like Droppable.js) or we force them to do DOM math to figure out what element of theirs our draggable landed on.

subdavis:
Bump

zackgalbreath:
Thanks for working on this, Brandon! I can confirm our customer is still interested in drag & drop from a girder file browser. I'll give this a try & let you know how it goes for me.

zachmullen:
Thanks for the explanation. In order to maintain API stability for the future in case we do decide to allow multiple drag and drop, I suggest that when we trigger an event, we send an Array of rows rather than a single row. For this implementation, it would always be an array of size 1.

subdavis:
Sure, 3da5160 adds this change.

PR #108, opened 2019-04-05
Add TOTP support

### LINKED ISSUE #107: Support login with 2FA
The login widget needs to support users with 2FA enabled.

subdavis:
Screenshot_20190405_151407

The only issue I see here is the mask `######` which will break if the OTP length changes.  It would be nice to include that information somehow in the authentication response if in the future users can configure the strength of token they want.

resolves #107

### DIFF
--- a/src/components/Authentication/Login.vue
+++ b/src/components/Authentication/Login.vue
@@ -10,6 +10,7 @@
   v-container
     v-form(@submit.prevent="login", ref="login")
       v-text-field(
+          v-if="!otpFormVisible",
           v-model="username",
           label="Username or e-mail",
           autofocus,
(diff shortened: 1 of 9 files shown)


jeffbaumes:
Cool! Since I've never heard of the phrase "TOTP token", maybe have it say something more user-friendly like "One-time password"?

zachmullen:
Should we add a prop for showing the OTP field from the get-go if downstreams know in advance that OTP is required? That way only one login request would need to be made, and users would know they have to fill out both fields.

subdavis:
@zachmullen I think that's a reasonable override.  I'll name that `forceOtp` with default `false`.

subdavis:
If upstream forces otp, it becomes a required field.

subdavis:
Added optional force prop.  Changed `TOTP Token` to `Authentication Code`.

src/components/Authentication/Login.vue
```diff
@@ -10,25 +10,35 @@
...
           label="Password",
           :rules="nonEmptyRules",
           prepend-icon="$vuetify.icons.lock")
-      v-layout(row)
+      v-text-field(
+          v-if="otpFormVisible || forceOtp",
+          v-model="otp",
+          type="text",
+          mask="######",
+          label="Authentication code",
```
brianhelba: "Two-factor authentication code" is likely the most obvious name for the end user.

Consistently referring to this internally as a OTP is most accurate, however.

src/components/Authentication/Login.vue
```diff
@@ -10,25 +10,35 @@
...
           type="password",
           label="Password",
           :rules="nonEmptyRules",
           prepend-icon="$vuetify.icons.lock")
-      v-layout(row)
+      v-text-field(
+          v-if="otpFormVisible || forceOtp",
+          v-model="otp",
+          type="text",
+          mask="######",
```
brianhelba: It's safe to assume this will be 6 digits for the foreseeable future. Nearly all TOTP clients (mobile phone apps, etc.) assume a 6 digit key (with a few supporting 8), so it would break support for many clients to change the OTP length.

PR #109, opened 2019-04-08
Improve handling of cookies with CORS requests

brianhelba:
* Rename the RestClient `cors` parameter to `setLocalCookie`
  * This is a more accurate description of its behavior.
* Add a `withCredentials` parameter to RestClient
  * This will send all XHR requests with this option.

### DIFF
--- a/src/rest.js
+++ b/src/rest.js
@@ -15,7 +15,7 @@ export default class RestClient extends Vue {
     apiRoot = '/api/v1',
     token = cookies.get('girderToken'),
     axios = axios_.create(),
-    cors = true,
+    setLocalCookie = true,
   } = {}) {
     super({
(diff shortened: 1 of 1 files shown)


zachmullen:
I'm fine with this change. We should definitely add a docstring to this constructor that describes all these options.

brianhelba:
@zachmullen Done.

brianhelba:
@zachmullen I'm actually fine with this, as long as I have a way to do it downstream.

To be clear, I believe it would be:
```javascript
girderRest.interceptors.request.use(config => ({
  ...config,
  withCredentials: true,
}));
```

brianhelba:
I've updated the PR accordingly

PR #110, opened 2019-04-08
Rewrite the README user documentation

brianhelba:
This resolves some incorrect or missing information, while refocusing the user documentation as a quickstart guide for typical usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Girder web component library
 
-Try the [demo app](https://girder.github.io/girder_web_components/).  It works with [data.kitware.com](https://data.kitware.com/).
+Try the [demo app](https://girder.github.io/girder_web_components/).
+It works with [data.kitware.com](https://data.kitware.com/).
 
 ## For users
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,125 +13,145 @@ npm install @girder/components
...
 ```
 
-> **Note:** When importing components from the source tree, you should import
-> them from `index.js` rather than importing the `.vue` files yourself, as the
-> latter is prone to break if files get moved in future releases. For example:
-> ```javascript
->  import { Upload as GirderUpload } from '@girder/components/src/components';  // Good
->  import GirderUpload from '@girder/components/src/components/Upload.vue'; // Unsafe -- may move in future
-> ```
-> Files and symbols that do not appear in an index.js should be considered private and it
```
zachmullen: Are we intentionally getting rid of this normative bit about not using symbols unless they're exported all the way to the top level of the UMD module?

brianhelba: It's now simplified above:
> Components should be imported from `@girder/components/src/components`, as this path will be stable across releases.

subdavis: I think that may be a little vague.  `@girder/components/src/components` could be interpreted as a base path, and when I'm looking through READMEs, I usually skim over sentences and read code examples.

I'd prefer to keep this section as-is.

README.md
```diff
@@ -12,125 +13,145 @@ npm install @girder/components
 yarn add @girder/components
 ```
 
-### Use as a prebuilt UMD module
+[VueCLI](https://cli.vuejs.org/) is required for building applications with Girder web components.
+However, a few additional packages must still be manually installed:
+```bash
+npm install -D sass-loader node-sass pug-plain-loader pug
+# or
+yarn add -D sass-loader node-sass pug-plain-loader pug
```
subdavis: I wish there were a better way to communicate these requirements within the framework of package.json.

I still sorta think we could (ab)use `peerDependencies` for these since girder-web-components will never have a use for the traditional `peerDependency` use case.

README.md
```diff
@@ -12,125 +13,145 @@ npm install @girder/components
...
-<script src="https://unpkg.com/vue/dist/vue.min.js"></script>
-<script src="https://unpkg.com/@girder/components"></script>
-<link rel="stylesheet" href="https://unpkg.com/@girder/components/dist/girder.css">
-```
+Encapsulating the configuration in another file (typically `src/plugins/girder.js`) is a good practice:
 
-This will expose all the library's components under the global variable `girder`, e.g.
-`girder.components.Upload`.
+```javascript
+import Vue from 'vue';
```
subdavis:
```suggestion
// src/plugins/girder.js

import Vue from 'vue';
```

I always really appreciate when documentation with multiple files in-line comment their hypothetical paths at the top.

README.md
```diff
@@ -12,125 +13,145 @@ npm install @girder/components
...
+export default GirderProvider;
+```
+
+Reference the configuration from your application initialization (typically `src/main.js`):
 
 ```javascript
-import * as girder from '@girder/components';
-// apiRoot should point to your girder instance
-const restClient = new girder.RestClient({ apiRoot: 'http://localhost:8080/api/v1' });
+import GirderProvider from '@/plugins/girder';
```
subdavis:
```suggestion
// main.js

import GirderProvider from '@/plugins/girder';
```

README.md
subdavis: Probably not the right place for this discussion.  I just think this info is better communicated through package.json than README.md

README.md
brianhelba: Will fix.

README.md
brianhelba: Will fix.

README.md
brianhelba: I don't think so. This is abusing the fact that `peerDependencies` print warnings and not cause the build to fail when not satisfied, but `peerDependencies` are not mere informational suggestions, they have behavior which relates to how conflicting dependencies are resolved.

The long-term solution / "right answer" is for GWC to publish ES5-compatible module code, so downstreams are not burdened with having to apply a Pug / Sass build process just to import GWC.

README.md
brianhelba: I'll pull the code examples back into this section. I agree, we don't want to make the text longer.

README.md
zachmullen: > The long-term solution / "right answer" is for GWC to publish ES5-compatible module code

That's what the UMD module is for, are you also suggesting we should publish source files transpiled to ES5? Are there good tools to make that easy?

brianhelba: You can't do tree-shaking downstream from a monolithic UMD file, and you also don't want to push any potentially-hard-to-build features (anything from Pug usage to basic ES6 features which would need to go though Babel) downstream either. So yes, the solution is to publish ES5-compatible modules on a per-file basis.

Here's some reading:
* https://github.com/rollup/rollup/wiki/pkg.module
* https://webpack.js.org/guides/author-libraries/#final-steps
* https://medium.com/webpack/webpack-and-rollup-the-same-but-different-a41ad427058c

brianhelba:
I'll wait for @zachmullen to chime in before merging.

README.md
```diff
@@ -12,125 +13,153 @@ npm install @girder/components
...
-    ...
+  components: {
+    GirderAuthentication,
+  },
+
+  // Injecting is not necessary to use the component,
+  // but is required to access the results of API calls
+  inject: ['girderRest'],
+  created() {
+    this.girderRest.$on('login', this.onLogin);
```
zachmullen: I realize this is just a toy example, but I fear people may construe this as a normative pattern for using RestClient.

The `user` property of RestClient is reactive, and is meant to be used as such. I'd suggest we come up with a contrived example of using it reactively, such as 

```pug
div Hello {{ girderRest.user ? girderRest.user.login || "anonymous" }}
```

brianhelba: @zachmullen I updated the example.

Issue #103, opened 2019-03-25
Minor bugs and nits


subdavis:
* Databrowser should properly flow to the size of its container
* use new slot syntax (2.6) for GWC.  Upgrade vuetify version.
* key slotted actions in data browser.

Might break these into individual issues.  Thanks, @matthewma7

matthewma7:
Thank you @subdavis for collect these items from my use case.
I think there is one more item might be useful

- Send internal rows as an event from the DataBrowser

subdavis:
Moved this additional feature to #104

zachmullen:
I took the liberty of modifying the original description comment to be a list of checkboxes, to make this easier to track and close out. If any of them are already done, please feel free to add checks.

PR #112, opened 2019-04-09
Make entire row clickable when selectEnabled = false, emit rowClick

subdavis:
When `selectEnabled == false`, make the entire row clickable.

More changes for #103


matthewma7:
Tried this out just then. Thanks! @subdavis

PR #114, opened 2019-04-10
Import Vuetify as a module, reducing size

brianhelba:
See: https://vuetifyjs.com/en/framework/a-la-carte#importing-components

Since this allows Webpack's tree-shaking to occur, this reduces the size of the resulting UMD library and demo app by ~40%.

Here's the specific results of applying this change:
Before: `build`
```
  File                      Size                     Gzipped

  dist/girder.umd.min.js    645.97 KiB               159.85 KiB
  dist/girder.umd.js        1490.58 KiB              284.00 KiB
  dist/girder.common.js     1490.12 KiB              283.83 KiB
  dist/girder.css           358.94 KiB               53.17 KiB
```

Before: `build:demo`
```
  File                                    Size              Gzipped

  _site/js/chunk-vendors.09a74c68.js      657.34 KiB        168.95 KiB
  _site/js/app.340b2dbd.js                56.74 KiB         14.27 KiB
  _site/css/chunk-vendors.82369c17.css    355.21 KiB        52.33 KiB
  _site/css/app.1ea9639a.css              3.82 KiB          1.17 KiB
```

After: `build`
```
  File                      Size                     Gzipped

  dist/girder.umd.min.js    368.81 KiB               113.88 KiB
  dist/girder.umd.js        948.40 KiB               222.27 KiB
  dist/girder.common.js     947.93 KiB               222.08 KiB
  dist/girder.css           395.72 KiB               59.05 KiB
```

After: `build:demo`
```
  File                                    Size              Gzipped

  _site/js/chunk-vendors.0ad10a31.js      404.42 KiB        131.42 KiB
  _site/js/app.1f8e3ccb.js                59.07 KiB         15.02 KiB
  _site/css/chunk-vendors.ca28d749.css    392.14 KiB        58.31 KiB
  _site/css/app.1ea9639a.css              3.82 KiB          1.17 KiB
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,9 @@ yarn add @girder/components
 [VueCLI](https://cli.vuejs.org/) is required for building applications with Girder web components.
 However, a few additional packages must still be manually installed:
 ```bash
-npm install -D sass-loader node-sass pug-plain-loader pug
+npm install -D sass-loader node-sass pug-plain-loader pug stylus-loader stylus vue-cli-plugin-vuetify vuetify-loader
 # or
-yarn add -D sass-loader node-sass pug-plain-loader pug
+yarn add -D sass-loader node-sass pug-plain-loader pug stylus-loader stylus vue-cli-plugin-vuetify vuetify-loader
 ```
 
 ### Initialization
--- a/package.json
+++ b/package.json
@@ -20,7 +20,9 @@
     "qs": "^6.5.2",
     "vue": "^2.5.21",
     "vue-async-computed": "^3.4.1",
-    "vuetify": "^1.3.15"
+    "vue-cli-plugin-vuetify": "^0.5.0",
+    "vuetify": "^1.3.15",
+    "vuetify-loader": "^1.2.1"
   },
   "devDependencies": {
     "@vue/cli-plugin-babel": "^3.0.0",
@@ -37,6 +39,8 @@
     "sass-loader": "^7.0.1",
     "stylelint": "^9.6.0",
     "stylelint-config-standard": "^18.2.0",
+    "stylus": "^0.54.5",
+    "stylus-loader": "^3.0.2",
     "vue-cli-plugin-pug": "^1.0.7",
     "vue-template-compiler": "^2.5.16"
   },
@@ -80,7 +84,9 @@
     },
     "moduleNameMapper": {
       "^@/(.*)$": "<rootDir>/src/$1",
-      "\\.css$": "<rootDir>/tests/mocks/stub.js"
+      "\\.css$": "<rootDir>/tests/mocks/stub.js",
+      "^vuetify/lib$": "vuetify",
+      "^vuetify/lib/(.*)": "vuetify/es5/$1"
     },
     "snapshotSerializers": [
       "jest-serializer-vue"
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,5 @@
 import VueAsyncComputed from 'vue-async-computed';
-import Vuetify from 'vuetify';
+import Vuetify from 'vuetify/lib';
 import * as components from './components';
 import * as constants from './constants';
 import RestClient from './rest';
--- a/src/utils/vuetifyConfig.js
+++ b/src/utils/vuetifyConfig.js
@@ -4,7 +4,7 @@
  * installing the GirderVue plugin.
  */
 
-import colors from 'vuetify/es5/util/colors';
+import colors from 'vuetify/lib/util/colors';
 
 export default {
   theme: {
--- a/yarn.lock
+++ b/yarn.lock
@@ -2687,6 +2687,11 @@ css-loader@^1.0.1:
     postcss-value-parser "^3.3.0"
     source-list-map "^2.0.0"
 
+css-parse@1.7.x:
+  version "1.7.0"
+  resolved "https://registry.yarnpkg.com/css-parse/-/css-parse-1.7.0.tgz#321f6cf73782a6ff751111390fc05e2c657d8c9b"
+  integrity sha1-Mh9s9zeCpv91ERE5D8BeLGV9jJs=
+
 css-select-base-adapter@~0.1.0:
   version "0.1.1"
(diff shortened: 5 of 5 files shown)


zachmullen:
In case anyone didn't know, if you run `yarn build --report`, it will generate a really nice treemap visualization of your bundle. Here is a supplemental visualization to demonstrate the effect of this change. In the below images, the parts of the bundle that represent Vuetify are shaded red. Note that these images are specific to the library target `girder.umd.min.js`.

**master**
Screen Shot 2019-04-10 at 1 48 16 PM

**9fd9695**

Screen Shot 2019-04-10 at 1 47 40 PM

zachmullen:
Turns out those images don't paint a good picture due to [this issue](https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/161). The entry-lib.js has absorbed the tree-shaking results.

subdavis:
This is certainly a useful change.  I actually think the major benefit is not the size reduction, but a nice stable example of how to do this in other consumer applications.

Nobody really uses UMD and demo doesn't matter, but this sets a good example for how to consume vuetify elsewhere.

That being the case, could we also [drop the color pack](https://vuetifyjs.com/en/framework/colors#stylus-color-pack) ?  There are instances in the code where I used `grey` when I should have used `secondary` so that should be a different PR.

subdavis:
The only issue I can see:

If I do this in my app (copied from README)

```javascript
import Vuetify from 'vuetify';
//...
Vue.use(Vuetify, appVuetifyConfig);
Vue.use(Girder);
```

Wouldn't this cause some naive sap (like me) to end up rolling in both the packaged version (in my app) and the tree-shaken bits from `Vue.use(Girder)`?

As a side note, I think that may be a bug in the README: the custom config should come _after_ `Vue.use(Girder)`, since the current ordering would cause any custom config to be overwritten.

Even better: we could have our `install` take an additional parameter, do the config merge from within that function, and avoid the issue of ordering all together.  Then we could prescribe that consumer apps never instantiate their own vuetify to avoid the double-packaging issue.

brianhelba:
> Nobody really uses UMD and demo doesn't matter, but this sets a good example for how to consume vuetify elsewhere.

Sure, but it is technically necessary for GWC to import Vuetify in a modular way for downstreams to receive any benefit.

> Wouldn't this cause some naive sap (like me) to end up rolling in both the packaged version (in my app) and the tree-shaken bits from `Vue.use(Girder)`?

Maybe, but we can document that. If some downstream users are not able to follow the consistently recommended practices from both GWC and Vuetify's docs, then their builds will just end up larger.

> As a side note, I think that may be a bug in the README: the custom config should come _after_ `Vue.use(Girder)`, since the current ordering would cause any custom config to be overwritten by the defaults.

We'll have to look inside the logic of Vuetify to see whether it locks in the first config given (for instance, this is definitely how modifying Vuetify's Stylus variables works) or whether it overwrites with the most recent. I don't know.

> Even better: we could have our `install` take the override config as an optional additional parameter, do the config merge from within that function, and avoid the issue of ordering all together. Then we could prescribe that consumer apps never instantiate their own Vuetify to avoid the double-packaging issue.

-1 From a developer experience perspective, I don't like that importing GWC requires I cede all control of Vuetify to go though GWC. A set of optional parameters would be a better compromise. Remember, the only hard case is when we are dealing with downstreams that want their own custom config too.

subdavis:
> Maybe, but we can document that.

Sure, just a documentation of that nuance would be sufficient.

> We'll have to look inside the logic of Vuetify

[Looks like an overwrite](https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/Vuetify/index.ts#L31-L48), and behaves like an overwrite from my local testing.

> I don't like that importing GWC requires I cede all control of Vuetify to go though GWC

What sort of control is lost?

matthewma7:
I just realize I am relying on GWC to install Vuetify in a consumer app of GWC. I am wondering how it would work if the consumer app of GWC also uses Vuetify.

subdavis:
> I just realize I am relying on GWC to install Vuetify in a consumer app of GWC. I am wondering how it would work if the consumer app of GWC also uses Vuetify.

Pretty sure you wouldn't notice a difference.  I rely on GWC to install vuetify in a couple of projects.

brianhelba:
> > I just realize I am relying on GWC to install Vuetify in a consumer app of GWC. I am wondering how it would work if the consumer app of GWC also uses Vuetify.
> 
> Pretty sure you wouldn't notice a difference. I rely on GWC to install vuetify in a couple of projects.

@matthewma7 @subdavis I'd disagree; attempting to rely on transitive dependencies is an anti-pattern. While it's possibly to leverage how npm often hoists non-conflicted packages up to the top level `node_modules`, this behavior is very fragile and not guaranteed to occur. All packages which are explicitly imported within the JS of a package really must be listed as top-level `dependencies` or `peerDependencies` of that package.

brianhelba:
@subdavis @zachmullen Ping; do you want this PR?

matthewma7:
Hi @brianhelba I mean the Vue plugin install part instead of package.json. 
Currently, I am not calling `Vue.install(Vuetify)` in my application because GWC did it.
Currently, it probably won't do anything if I do call it because I heard Vuetify attach something to the global Vue object to track its install state. But with this change, I don't know if I need or can call `Vue.install(Vuetify)`.

zachmullen:
> do you want this PR?

Yes, but admittedly I don't fully understand the discussions about how this affects downstream use of Vuetify. Is there an executive summary? Just trying to make sure this doesn't cause any harm.

brianhelba:
To elaborate, I think @matthewma7 brings up really good points, but they're out of scope of the changes of this PR. I hope they can be condensed into an issue report, which can be followed up independently.

matthewma7:
@brianhelba I can give the use case I described a try after this is merged.

matthewma7:
Update: 
With this change, the downstream app that relies on GWC to install Vuetify breaks. 

- import Vuetify and `Vue.use(Vuetify)` before `Vue.use(Girder)` fixed the issue
- but when trying to do the a [la carte Vuetify](https://vuetifyjs.com/en/framework/a-la-carte) in the downstream app itself, the theme defined by GWC doesn't work correctly, no matter what order of Vue.use()

PR #116, opened 2019-04-11
Upload class optional members

matthewma7:
This PR includes a few improvements.

* Add back `preUpload`, `postUpload` callbacks. Because (1) The demo app is using these callback, the demo app is broken and won't refresh the file list after upload at this moment. (2) The context of these callbacks is the parent Vue component of Upload.vue. the context `afterUpload`, and `beforeUpload` is the upload class. Without hacking, It's way easier to consume the callback in the parent component with these callbacks. (3) `afterUpload`, and `beforeUpload` is for the individual file. `preUpload`, `postUpload` is for all of the uploads. Take the demo app as an example. It's meant to refresh after the whole upload process. Again, without hacking, these are easier to work with. 

* This line https://github.com/girder/girder_web_components/compare/upload_class_optional_members?expand=1#diff-82311ba10aa7e62082a059e895af2774R168 make progress bar progress if the custom uploader doesn't report progress. 
image

* This line https://github.com/girder/girder_web_components/compare/upload_class_optional_members?expand=1#diff-82311ba10aa7e62082a059e895af2774R135 make it show loading state if the custom uploader doesn't report its uploading. Same image above.

* https://github.com/girder/girder_web_components/compare/upload_class_optional_members?expand=1#diff-82311ba10aa7e62082a059e895af2774R112 and next line make the property responsive, so it's easier to work with when changing their values. (No need to use Object.assign({}, existing) to change the object all together) These also let people know what properties are valid beforehand without tracing the code. 

* https://github.com/girder/girder_web_components/compare/upload_class_optional_members?expand=1#diff-8af87c0ce07db8082e10e648f84bda61L34 is not used

* By adding the conditionals, the custom uploaders literally only need to provide start(), Otherwise, a custom uploader needs to provide all the callbacks itself or subclass upload.js, either way, it also needs to provide a resume(), even if the resume() simply calls start(). If we don't like the conditional, we should provide an 'abstract class' that a custom upload should inherit from. 

I find these improvements are useful because I have an app that uploads numerous text files, which is 1, doesn't go to upload endpoint (calls a custom endpoint) 2, small but many, so no need to report individual process, but need overall progress. and if failed due to a network issue, no need to resume() and just start from the beginning.

### DIFF
--- a/src/components/Upload.vue
+++ b/src/components/Upload.vue
@@ -53,6 +53,14 @@ export default {
       default: true,
       type: Boolean,
     },
+    preUpload: {
+      default: () => {},
+      type: Function,
+    },
+    postUpload: {
+      default: () => {},
+      type: Function,
+    },
     uploadCls: {
       default: Upload,
       type: Function,
@@ -113,7 +121,7 @@ export default {
       const results = [];
       this.uploading = true;
       this.errorMessage = null;
-
+      await this.preUpload();
       for (let i = 0; i < this.files.length; i += 1) {
         const file = this.files[i];
         if (file.status === 'done') {
@@ -180,6 +188,7 @@ export default {
           }
         }
       }
+      await this.postUpload();
       this.uploading = false;
       this.files = [];
       this.$emit('done', results);
--- a/src/components/Upload.vue
+++ b/src/components/Upload.vue
@@ -111,6 +111,7 @@ export default {
         progress: {
           indeterminate: false,
           current: 0,
+          size: file.size,
         },
         upload: null,
         result: null,
--- a/src/components/Upload.vue
+++ b/src/components/Upload.vue
@@ -145,29 +145,19 @@ export default {
                 progress,
               });
               // eslint-disable-next-line no-await-in-loop
-              await file.upload.beforeUpload({
-                current: i,
-                total: this.files.length,
-              });
+              await file.upload.beforeUpload();
               // eslint-disable-next-line no-await-in-loop
               file.result = await file.upload.start();
             }
             // eslint-disable-next-line no-await-in-loop
-            await file.upload.afterUpload({
-              current: i,
-              total: this.files.length,
-            });
+            await file.upload.afterUpload();
             delete file.upload;
             results.push(file.result);
             file.status = 'done';
             file.progress.current = file.file.size;
           } catch (error) {
             // eslint-disable-next-line no-await-in-loop
-            await file.upload.onError({
-              error,
-              current: i,
-              total: this.files.length,
-            });
+            await file.upload.onError(error);
 
             if (error.response) {
               this.errorMessage = error.response.data.message || 'An error occurred during upload.';
--- a/src/components/Upload.vue
+++ b/src/components/Upload.vue
@@ -136,7 +136,7 @@ export default {
           try {
             if (file.upload) {
               // eslint-disable-next-line no-await-in-loop
-              file.result = await (file.upload.resume ? file.upload.resume() : file.upload.start());
+              file.result = await file.upload.resume();
             } else {
               // eslint-disable-next-line new-cap
               file.upload = new this.uploadCls(file.file, {
@@ -144,37 +144,30 @@ export default {
                 parent: this.dest,
                 progress,
               });
-              if (file.upload.beforeUpload) {
-                // eslint-disable-next-line no-await-in-loop
-                await file.upload.beforeUpload({
-                  current: i,
-                  total: this.files.length,
-                });
-              }
-
               // eslint-disable-next-line no-await-in-loop
-              file.result = await file.upload.start();
-            }
-            if (file.upload.afterUpload) {
-              // eslint-disable-next-line no-await-in-loop
-              await file.upload.afterUpload({
+              await file.upload.beforeUpload({
                 current: i,
                 total: this.files.length,
               });
+              // eslint-disable-next-line no-await-in-loop
+              file.result = await file.upload.start();
             }
+            // eslint-disable-next-line no-await-in-loop
+            await file.upload.afterUpload({
+              current: i,
+              total: this.files.length,
+            });
             delete file.upload;
             results.push(file.result);
             file.status = 'done';
             file.progress.current = file.file.size;
           } catch (error) {
-            if (file.upload.onError) {
-              // eslint-disable-next-line no-await-in-loop
-              await file.upload.onError({
-                error,
-                current: i,
-                total: this.files.length,
-              });
-            }
+            // eslint-disable-next-line no-await-in-loop
+            await file.upload.onError({
+              error,
+              current: i,
+              total: this.files.length,
+            });
 
             if (error.response) {
               this.errorMessage = error.response.data.message || 'An error occurred during upload.';
--- a/src/components/Upload.vue
+++ b/src/components/Upload.vue
@@ -124,10 +124,11 @@ export default {
             Object.assign(file.progress, event);
           };
           file.status = 'uploading';
+          file.progress.indeterminate = true;
           try {
             if (file.upload) {
               // eslint-disable-next-line no-await-in-loop
-              file.result = await file.upload.resume();
+              file.result = await (file.upload.resume ? file.upload.resume() : file.upload.start());
             } else {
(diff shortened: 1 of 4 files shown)


src/components/Upload.vue
```diff
@@ -110,51 +121,60 @@ export default {
...
+      await this.preUpload();
       for (let i = 0; i < this.files.length; i += 1) {
         const file = this.files[i];
         if (file.status === 'done') {
           // We are resuming, skip already completed files
           results.push(file.result);
         } else {
           const progress = (event) => {
-            file.progress = Object.assign({}, file.progress, event);
+            Object.assign(file.progress, event);
```
zachmullen: Haven't tested, does reactivity still work properly with this method?

matthewma7:
@zachmullen I edited my original post after I post it initially. quote myself
> If we don't like the conditional, we should provide an 'abstract class' that a custom upload should inherit from.

What do you think? I think an abstract class would make it clear for what is asked and what is optional.

zachmullen:
I'm fine with having an abstract upload class if it helps for developer clarity.

src/components/Upload.vue
matthewma7: Yes. Somehow Object.assign will still call the property setter, making it reactive.

src/components/Upload.vue
```diff
@@ -110,17 +121,18 @@ export default {
       const results = [];
       this.uploading = true;
       this.errorMessage = null;
-
+      await this.preUpload();
```
zachmullen: How do others feel about reintroducing these hooks? They are duplicative with the design of the Upload class hooks, so they violate the spirit of having only one right way to accomplish something. On the other hand, I see the appeal for these as it makes for less code when you only need an action before the first upload or after the last, which is a common requirement. I am not sure which choice yields greater overall simplicity for users.

src/utils/upload.js
```diff
@@ -23,15 +25,17 @@ export default class Upload {
...
     Object.assign(this, {
-      $rest, file, params, parent, progress, chunkLen, upload: null, offset: 0, behavior: null,
+      params, chunkLen, upload: null, offset: 0, behavior: null,
     });
   }
 
   async _sendChunks() {
     const onUploadProgress = e => this.progress({
       current: this.offset + e.loaded,
-      total: this.file.size,
```
zachmullen: What's the rationale for not passing this to the callback? I realize the receiver almost certainly has access to the total, but it seems helpful (or at least harmless) to also include it here.

src/utils/UploadBase.js
```diff
@@ -0,0 +1,75 @@
...
+   * The abstract base class of a single file uploader.
+   * @abstract
+   * @param file {File | Blob} the file to upload
+   * @param opts {Object} upload options.
+   * @param opts.$rest {Object} an axios instance used for communicating with Girder.
+   * @param opts.parent {Object} upload destination. Must have ``_id`` and ``_modelType``.
+   * @param opts.progress {Function} A progress callback for the upload. It can take an Object
+   *   argument with either ``"indeterminate": true``, or numeric ``current`` fields.
+   */
+  constructor(
```
zachmullen: Since this is abstract, it seems like we should omit the constructor altogether. None of the properties assigned here in the constructor are referenced in the rest of the abstract class, meaning they can be pushed down as details of the concrete subclasses. (It's prototypal inheritance so there's no hard and fast right answer, but default/implicit ctor here is what feels right to me.)

zachmullen: If we do want to keep this to provide a kind of documentation of what parameters the ctor will receive (which is reasonable), we should not set the `upload`, `offset`, or `behavior` fields in the abstract ctor, as they are not passed to us.

src/components/Upload.vue
matthewma7: Allow me to talk in detail my point. I think they are not the same thing, one is for before and after each file and belongs to the uploader and another one is before and after all file upload and belongs to Upload.vue. They are both equally important, there is no straight forward way to achieve one's feature with the other one. 
Let's take [what the demo app is doing with postUpload](https://github.com/girder/girder_web_components/blob/master/demo/App.vue#L144) as an example, the steps I can think of is, first create a Class A that extends the built-in upload.js, then attached the postUpload() to the prototype of the Class A, then bind its context to `this` of App.vue, with arrow function or something. then, debounce the call to the function to only trigger after a certain delay, but it's still not deterministic, since one upload could take a long time, you never know when is the last one. All of these hacking can be subtituded with a simple postUpload() on Upload.vue 
Also, these preUpload() and postUpload() is parallel to the [postUpsert()](https://github.com/girder/girder_web_components/blob/master/demo/App.vue#L151)

zachmullen: It's true that this depends on our current status quo where all uploads are done in sequence rather than parallel. I anticipate that we could add a parallel option to the Upload.vue component, which would break this assumption. In light of that, I now agree with you that these are necessary.

Since these will now become the one true way of doing "before all" and "after all" hooks, we should simply remove the `current` and `total` parameters that are passed to the upload class hooks, since we don't see any other utility for them at this time.

src/utils/upload.js
matthewma7: I first noticed this is because the total is not being used in Upload.vue, then I realize Upload.vue is [aware of the size itself](https://github.com/girder/girder_web_components/blob/master/src/components/Upload.vue#L88). Because it is Upload.vue that passed the file to upload.js to upload, It's natural that Upload.vue is aware of things like file size, name. I think file size is something given to upload.js, not something it is able to change or responsible to report.

src/utils/UploadBase.js
matthewma7: Oh, yes. upload, offset, and behavior should not be mentioned in the Base class constructor.

src/components/Upload.vue
matthewma7: Sounds good. I'll remove related code.

src/utils/upload.js
zachmullen: We should also consider the fact that the upload class could be used in other contexts besides the Upload.vue component. Callers will likely always know the size, but it makes for a nice API to include the total IMO, so you can have a callback like:

```javascript
({ current, total }) => { this.percentDone = 100 * current / total; }
```

I could be in the minority here, though. What do others think? @subdavis @brianhelba

subdavis: I haven't tried to do anything with that callback, but I agree that the API is nice to have.  Even when `Upload.vue` is the source of the data, the handler could be somewhere else.   I wouldn't want to reach into that component's private data to get this info.

subdavis:
@matthewma7 thank you for putting this PR together.  I want to review it more thoroughly, I think I'll be able to this afternoon.  I also think there are some valuable parts here, and you've convinced me that the prop function callbacks are valuable.

matthewma7:
@subdavis Cool, glad I can make my reasoning clear.

src/utils/upload.js
matthewma7: Make sense, but can I take the liberty to change the name to size? Because size is more explicit than total and the word total has been used in Upload.vue for real total.

zachmullen: Fine by me.

zachmullen:
LGTM, @subdavis can add the explicit review later.

src/components/Upload.vue
```diff
@@ -133,28 +146,19 @@ export default {
...
             // eslint-disable-next-line no-await-in-loop
-            await file.upload.afterUpload({
-              current: i,
-              total: this.files.length,
-            });
+            await file.upload.afterUpload();
             delete file.upload;
             results.push(file.result);
             file.status = 'done';
+            file.progress.current = file.file.size;
```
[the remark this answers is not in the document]
matthewma7: This line makes the overall progress bar progress if the uploader class doesn't report progress by calling the progress callback.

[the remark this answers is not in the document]
subdavis: What's the point of that, since it would jump from `indeterminate` (spinning) to `done` (checkmark) anyway?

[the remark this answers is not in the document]
zachmullen: It essentially means there's sensible behavior for the upload even if you don't do anything at all with progress in your subclass.

PR #118, opened 2019-04-18
Internalize helper variable

matthewma7:
This PR internalize the DebounceCounter helper variable to an individual instance of the Search component. Even though I don't see we could run into issues in a short time with the current implementation, but this change prevents potential weird behavior and anti-pattern and help me monkey patch this Search in a downstream application.

### DIFF
--- a/src/components/Search.vue
+++ b/src/components/Search.vue
@@ -2,8 +2,6 @@
 <script>
 import { DebounceCounter } from '../utils';
 
-const counter = new DebounceCounter();
-
 export default {
   props: {
     maxQuickResults: {
@@ -26,7 +24,7 @@ export default {
   },
   computed: {
     loading() {
-      return counter.flag;
+      return this.counter.flag;
     },
     quickResults() {
       return this.searchResults.slice(0, this.maxQuickResults);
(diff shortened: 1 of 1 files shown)


subdavis:
I'm sorry, I don't understand what this change fixes or enables.  Could you reference the upstream where this is relevant? Thanks.

zachmullen:
This does seem like it should be one per component instance, even if we never run into issues with it practically. Unless someone can provide a rationale for it being globally shared, I think we should make this change.

matthewma7:
@subdavis Current if we create two <Search /> they will use the same instance of `counter`, which I believe is not what we meant to do. 
For my downstream code specifically, if the counter is defined with a function scope, I can't get a reference to it to do my monkey patching (Extending the Search to do some additional application specific search).

zachmullen:
> if the counter is defined with a function scope, I can't get a reference to it to do my monkey patching

Is this just in a testing environment, or are you monkey patching the counter at application runtime? If the latter, why do you need to do that?

matthewma7:
@zachmullen I am monkey patching at runtime, the reason is that I am making another custom request and I want to wait for both requests to finish, so I am calling the counter.inc() in my custom code as well.

matthewma7:
Thanks.

PR #120, opened 2019-04-23
Drag example

subdavis:
For @zackgalbreath 

Note that `dragend()` is introduced in this PR, so that handler won't work.

This PR adds `application/x-girder-items` data to `DragEvent.dataTransfer`.  If your app just has 1 dropzone, you can listen to `dragend` and get the full js object list of items dragged, but if you have > 1 dropzone, that event fires from the source, so it's not possible to determine what the destination target was and you have to listen to the actual `drop` event, and there's no way to get a javascript object from source to destination event.

You can, however, pass strings.  I think I'll have to add `_modelType` to the JSON serialized string, which is a little gross, but would make the drop event contain all the data you might need to fetch that item/folder/whatever from girder.

### DIFF
--- a/demo/App.vue
+++ b/demo/App.vue
@@ -57,7 +57,14 @@ v-app.app
               :new-folder-enabled="newFolderEnabled",
               @click:newitem="uploader = true",
               @click:newfolder="newFolder = true",
-              @selection-changed="selected = $event")
+              @selection-changed="selected = $event",
+              @dragend="dragend")
+        v-card.mt-3(v-if="dragEnabled", @dragenter="prev", @dragover="prev", @drop="drop")
+          v-card-title
+            h3.headline Drop Zone
+          v-card-text
+            p(v-if="!(dropped.length)") Drag a row here to see results
+            p(v-else, v-for="drop in dropped") {{ drop.item._modelType }} {{ drop.item.name }} {{ drop.item.size }} 
 </template>
 
 <script>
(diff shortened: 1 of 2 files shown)


demo/App.vue
```diff
@@ -152,12 +154,11 @@ export default {
...
       event.preventDefault();
     },
-    dragend({ items, event }) {
+    dragend({ items }) {
       this.dropped = items;
     },
     drop(event) {
-      const identifiers = event.dataTransfer.getData('application/x-girder-items');
-      console.log(identifiers);
+      this.droppedIDs = event.dataTransfer.getData('application/x-girder-items');
```
zackgalbreath: Thanks for the example, Brandon! I can confirm that this worked for me too.

cjh1:
@subdavis Can we get this drag event merged into master in some form? We have an app that is hanging out on the SHA and would like to update to the current release.

subdavis:
@cjh1 this PR is quite old and will need to be rebased to work.  We've upgraded several dependencies, including Vue, vuetify, and vuetify-loader.  Is your app compatible with https://github.com/girder/girder_web_components/blob/master/package.json ?

If so, I'm happy to update this PR and try to get things merged.

cjh1:
@subdavis We will be happy to make the updates necessary in our app. So if you can do that it would be great.

cjh1:
@bnmajor We will need to adapt to the version changes here: https://github.com/girder/girder_web_components/blob/master/package.json

subdavis:
Closing in favor of #243

Issue #119, opened 2019-04-23
Add ability to browse across user folders and collections to DataBrowser


matthewma7:
Currently, the DataBrowser requires an initial location and is not able to browse across user folders and collections. A new prop can be added to allow the DataBrowser to be able to browse everything the user has access to across user folders and collections.
This would also make location prop optional.

subdavis:
@matthewma7 thanks for logging this issue.  I've been missing this feature as well, so I think it's time someone added it.

Could we briefly discuss how this implementation will work?  Are collections / users going to behave as "folders" to the Data Browser such that navigation within the browser works the same?  Will the list of users be browse-able? How will this affect the breadcrumb?  I think it would be good to have the icon (currently a globe) represent user/collection, but I don't know how to allow the user to "ascend" from user/collection list to the very root which would just be:

```
+---------------------------------------------------------+
| Breadcrumb: What goes here?                             |
+---------------------------------------------------------+
|                                                         |
| X Users                                         100MB   |
|                                                         |
| X Collections                                   3.4GB   |
|                                                         |
+---------------------------------------------------------+
```

Maybe @jtomeck should be present and sign off on whatever we decide.  Maybe the two of you could talk about it.

I think I would enjoy adding this feature, but I'm a bit busy at the moment so if you have some allocation to spend, that would be awesome.  Thanks!

matthewma7:
Hi @subdavis, these are great questions and suggestions. 
> Are collections / users going to behave as "folders" to the Data Browser such that navigation within the browser works the same

I think yes. 
> Will the list of users be browse-able?

I didn't fully understand the question, but I think yes?

> Will the list of users be browse-able? How will this affect the breadcrumb? I think it would be good to have the icon (currently a globe) represent user/collection

I am thinking have the globe goto the page you have shown above, and [globe icon]/users and [globe icon]/collections to represent users and collections respectively. After that, the rest of breadcrumb remains what it is. so, [globe icon]/users/matthewma, [globe icon]/collections/collectionA

And this does make the breadcrumb long and we may not want to limit the user, so additional prop with default true or false will be provided.  

I definitely could have missed something here, other thoughts?

subdavis:
I think we're on the same page.

To conserve breadcrumb space, I think it would be nice to do the following:

For users regular folders (notice that `users / username` has been condensed):

`🌎 / 😃 username / folder_name`

For collection regular folders:

`🌎 / 📚 collection_name / folder_name`

For users:

[...]

PR #121, opened 2019-04-26
Add support for browsing collection and user with DataBrowser

matthewma7:
This PR adds support of user and collection browsing to the DataBrowser and  Breadcrumb components. This PR should meet all requirement of the issue and would resolve https://github.com/girder/girder_web_components/issues/119.
This requires girder>=3.0.0a7.dev71

This PR introduces a concept of *Root* in terms of the location. concept Root includes 'root', 'users', 'collections'.
~The concept of location was using girder models of type user, collection, and folder. However, with the introduction Root, the location has been changed to has two properties, a `type` property, and an optional `id` property. 
The type can be 'root', 'users', 'collections', 'user', 'collection', or 'folder'. When the value is 'user', 'collection', or 'folder', it should has the id field.~
The location props will be set to {type: 'root'/'users'/'collections'} when the location is not a girder user, collection or folder.

This PR also makes location prop on DataBrowser optional, so the user of DataBrowser doesn't have to provide a location or make it sync.

The fact that Upsert-Folder and Upload dialog is not internalized in DataBrowser makes the App.vue of the demo a little bit awkward. And, also, I and Jeff think we should standardize boolean props, to make component easier to be used. I am thinking we could do that in following PRs.


matthewma7:
@jeffbaumes @subdavis @brianhelba Any feedback? 😉

src/components/DataBrowser.vue
```diff
@@ -19,8 +20,10 @@ export default {
     },
     location: {
       type: Object,
-      required: true,
-      validator: val => val._modelType && val._id,
+      default: () => ({
+        type: 'root',
+      }),
+      validator: getLocationValidator(true),
```
jeffbaumes: I kind of liked the symmetry of inputting a model object (with `_modelType` and `_id`) as a prop and receiving similar model objects on event callbacks. Could we keep the same API for the prop as before and just have special values of "root", "users", and "collections" for the `_modelType`, where the `_id` would be ignored in those cases?

matthewma7: Hm, I see the convenience of just passing a girder model to location and the DataBrowser will navigate to it, but I think putting _modelType and these virtual things "root", "users", and "collections" is not a good thing to do. How about the location prop could take either a location object `{location: 'root'}`, or a 'user', 'folder', 'collection' model, but for the sake of consistency and to avoid confusion, I think the 'update: location' should always report `location` object. (we can now not listen to it if we don't need to know the location.) Since the 'itemclick' and 'selection-changed' is the event for objects?

src/utils/getLocationValidator.js
```diff
@@ -0,0 +1,15 @@
+export default function (allowRoot) {
```
subdavis: Since this is intended to be  part of the public utilities api, please expose through `utils/index.js` and import from `utils/`

src/components/DataBrowser.vue
subdavis: I agree with Jeff, and would like to keep this interface the same.  It's quite nice for upstream to be able to pass responses directly from girder API requests into GWC.  I don't think consumers should have to do mapping to make these components happy.

jeffbaumes: @subdavis I've talked to @matthewma7 about this just now, and I'm ok either way and retract my opinion. Basically, @matthewma7's implementation is inconsistent with prior behavior and other model objects floating around in a Girder client app, but your suggestion is inconsistent with what "_modelType" means, in that there is no "root", etc. model in Girder, which may be confusing to people looking at REST endpoints and server code. Since I don't think we can win here, I'm ok with either approach.

jeffbaumes:
To capture some other discussion with @matthewma7 just now, the plan is to simplify the "home" navigation UI to just have a small home icon to the left of the globe icon that is visible if someone is logged in, that always goes to the user home.

src/components/DataBrowser.vue
matthewma7: Hi @subdavis, I just had an offline discussion with @jeffbaumes. He agreed that putting _modelType: 'root' is confusing for people who are not as familiar with girder. This location is purely a GWC concept and keeping it explicit make it easier to understand. But to make it more convenient, I would update to make the location prop be able to take a user, collection, folder model as input, so the user of the component doesn't have to do the mapping.

demo/App.vue
```diff
@@ -155,9 +156,9 @@ export default {
     },
     handleSearchSelect(item) {
       if (['user', 'folder'].indexOf(item._modelType) >= 0) {
-        this.browserLocation = item;
+        this.browserLocation = { type: item._modelType, id: item._id };
```
subdavis: per below comment: this is the sort of translation I think it would be best to avoid.

src/components/Breadcrumb.vue
```diff
@@ -1,10 +1,12 @@
 <script>
+import getLocationValidator from '../utils/getLocationValidator';
+
 export default {
   props: {
     location: {
       type: Object,
       required: true,
-      validator: val => val._modelType && val._id,
+      validator: getLocationValidator(true),
```
subdavis: I'm confused somewhat by this line.  Should I be able to set `noRoot: true` but still allow a location with root to validate?

src/components/Breadcrumb.vue
```diff
@@ -23,44 +29,71 @@ export default {
...
-          const { data } = await this.girderRest.get(`${_modelType}/${_id}`);
-          breadcrumb.root = this.extractCrumbData(data);
+        } else if (type === 'user' || type === 'collection') {
+          const { data } = await this.girderRest.get(`${type}/${id}`);
+          breadcrumb.unshift(this.extractCrumbData(data));
+        }
+        if (!this.noRoot) {
+          if (
+            type === 'users' ||
+            (user && breadcrumb.length && breadcrumb[0].type === 'user')
```
subdavis: Why is `user` part of this conditional?  If I'm not logged in and browsing a public file in some real user's space, won't this evaluate false?

src/components/Breadcrumb.vue
matthewma7: This is something Vue prop validator won't able to do. Vue validator cannot take other props as part of the input. So, I validate allowing root here and later validate in connection with the no-root prop here. https://github.com/girder/girder_web_components/blob/browse_user-folder_collection_with_data-browser/src/components/Breadcrumb.vue#L86-L88

src/components/Breadcrumb.vue
matthewma7: This is for the 'users' location that lists all users, recent changes of girder make it only authenticated users are able to list all users. So, if the user is not authenticated, hide the 'users' location from the breadcrumb.

src/components/Breadcrumb.vue
```diff
@@ -23,44 +29,71 @@ export default {
   },
   asyncComputed: {
     breadcrumb: {
-      default: { root: {}, path: [] },
+      default: [],
```
subdavis: Rather than the very long conditional block in the template,  I think much of this logic could be simplified by making the breadcrumb schema:

`{ filespace: enum('users', 'collections', null), path: [] }`

The default would be:

`{ filespace: null, path: [] }`

* "root" is static and determined by the value of noRoot
* "fileSpace" will always be one of `'users'`, `'collections'`, `null`
* "path" can have the exact same logic as before.

src/components/Breadcrumb.vue
subdavis: Got it, thanks.

src/components/DataBrowser.vue
subdavis: @matthewma7 sure, if we can hide the mapping such that this isn't a breaking change, I'm happy with that.  Could you make that change in both DataBrowser and Breadcrumb? ty.

src/components/Breadcrumb.vue
matthewma7: Honestly, the conditional in the template is long but not complex. so I don't think introducing new concepts is simplifying.

src/components/DataBrowser.vue
matthewma7: Yes. I will make both of them can take girder model as the initial value.

src/components/Breadcrumb.vue
subdavis: Fewer conditionals in the template improve testability.  There's enough branching going on in the template that if we moved it into the computed prop itself, it would be a lot easier to get coverage on.

matthewma7:
Added tests. This is ready for review.

subdavis:
Thanks Matthew.

I still have a few small comments:

Request changes:

* https://github.com/girder/girder_web_components/pull/121#discussion_r279065432
* Could we rename `getLocationValidator` to either `createLocationValidator` or `makeLocationValidator` to better indicate that it returns a function?  I think `create` would be most appropriate.

Discuss:

* I'm still unsure about the template conditionals.  While the statements are pretty simple, they do represent 5 new branch cases we are unable to get coverage on (or more accurately, it is significantly more brittle to get coverage on because we have to inspect the DOM).  I think I'd like input from at least one more person on [this thread](https://github.com/girder/girder_web_components/pull/121#pullrequestreview-231326031).

Thanks again for putting this PR together.

matthewma7:
Thanks, @subdavis for revisiting this. 
I think I already addressed [#121 (comment)](https://github.com/girder/girder_web_components/pull/121#discussion_r279065432)?
I agree, `create` sounds more accurate. 

About the conditional on the template, we can totally see other input on this thread. 
I personally am thinking since all of them are else if, they are essentially `switch` statement, which is not that branch complex. and should have been covered, (but not tested?), if you are concerned that they are not tested, do you think adding tests with [wrapper.find](https://vue-test-utils.vuejs.org/api/wrapper/#find) would solve that?

matthewma7:
Cool! thank you for the reviewing and approval. @subdavis 
I'll merge this in one hour.

PR #126, opened 2019-06-07
Alpha v4

zachmullen:


zachmullen:
Just published this so I could take advantage of the upload class features downstream

PR #127, opened 2019-07-08
Migrate Job listing components from other repo

matthewma7:
This ports the Job Listing components from https://github.com/Kitware/resonantgeo
The development ResonantGeo will focus on Geospatial components and won't have the Job Listing components.
However, at this moment, Vigilant is using the Job Listing component, so it is important to have it moved to here and merge this sooner rather than later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "axios": "^0.18.0",
     "js-cookie": "^2.2.0",
     "markdown-it": "^8.4.2",
+    "moment": "^2.24.0",
     "qs": "^6.5.2",
     "vue": "^2.5.21",
     "vue-async-computed": "^3.4.1",
--- a/src/components/job/FilterForm.vue
+++ b/src/components/job/FilterForm.vue
@@ -1,0 +1,86 @@
+<script>
+import * as status from './status';
+
+export default {
+  props: {
+    jobType: {
+      type: [String, Object],
+      default: null,
+    },
+    jobTypeList: {
+      type: Array,
+      default() {
+        return [];
+      },
+    },
+    status: {
+      type: [Number, Object],
+      default: null,
+    },
+    statusList: {
+      type: Array,
+      default() {
+        return [];
+      },
+    },
+  },
+  data() {
+    return {
+      fromDateMenu: false,
+      toDateMenu: false,
+    };
+  },
+  computed: {
+    statusItemList() {
+      return this.statusList
+        .map(status.getByValue)
+        .filter(s => s && s.text)
+        .sort((a, b) => {
+          if (a.text > b.text) {
+            return 1;
+          } else if (a.text < b.text) {
+            return -1;
+          }
+          return 0;
+        });
+    },
+  },
+};
+</script>
+
+<template lang="pug">
+v-card.job-filter(dark, color="primary darken-1")
+  v-container.py-2(fluid, grid-list-lg)
+    v-layout(row)
+      v-flex(sm12, d-flex)
+        h3.primary--text.text--lighten-2 Jobs
+    v-layout(row, justify-center)
+      v-flex(sm3, d-flex)
+        v-select(
+            label="Job Type",
+            :items="jobTypeList",
+            :value="jobType",
+            clearable,
+            color="white",
+            :menu-props="{'content-class':'girder-job-filter-form-menu'}",
+            dark,
+            dense,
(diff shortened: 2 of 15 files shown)


package.json
```diff
@@ -16,7 +16,9 @@
     "@mdi/font": "^3.5.95",
     "axios": "^0.18.0",
     "js-cookie": "^2.2.0",
+    "lodash": "^4.17.5",
```
zachmullen: I vote for removing the lodash dependency, at least from the runtime code. I think we can do without the deep cloning for now, and `sortBy` is easy enough to do with standard functions.

matthewma7: One of the dependency is already using lodash, and I think sooner or later we would need it so I added it.
How to remove lodash from runtime code?

zachmullen: If we're already bundling it into the runtime that's fine, guess it won't add any extra weight. Can we show a before and after treemap of the bundle to make sure?

src/components/job/FilterForm.vue
```diff
@@ -0,0 +1,162 @@
...
+            label="Job Type",
+            :items="jobTypeList",
+            :value="jobType",
+            clearable,
+            color="white",
+            :menu-props="{'content-class':'filter-form-menu'}",
+            dark,
+            dense,
+            @input="$emit('update:jobType', $event ? $event : null)")
+      // for when girder's api supports this sort of filtering
```
subdavis: Can we remove the dead code?

Maybe replace with a reference to where this code came from in another repository?

src/components/job/FilterForm.vue
```diff
@@ -0,0 +1,162 @@
...
+            clearable,
+            color="white",
+            :menu-props="{'content-class':'filter-form-menu'}",
+            dark,
+            dense,
+            @input="$emit('update:status', $event ? $event : null)")
+</template>
+
+<style lang="scss">
+.filter-form-menu .v-list__tile--active.theme--light {
```
subdavis: Could we give the global style a more specific name, like `.gwc-job-filter-form-menu` ?

matthewma7: Yes, I am thinking `girder-job-filter-form-menu`.

src/components/job/FilterForm.vue
matthewma7: This code was originally here before they are migrated to here, but if it's ok, I can remove them but leave them in as commit history track them. Let me know if I should keep them @jbeezley 😉

jbeezley: I would remove it.  I originally wrote a faithful reproduction of Jared's design thinking I might implement the filtering, but it never happened.

matthewma7:
I pushed fixes for the issues, please take another look.

package.json
matthewma7: Ah, turns out lodash was dependancy for only development packages and also the moment.js locale stuff is huge. 
Importing lodash as like this will help to reduce the size greatly `import cloneDeep from 'lodash/cloneDeep';`, 
So is that ok I import functions like this to set as an example? I think sooner or later we will find it better to use lodash function for something more complex than sortBy than roll our own.

matthewma7:
Updated, please take another look.

package.json
zachmullen: In the spirit of YAGNI, let's not pull in lodash unless/until we really do need it. The cloneDeep actually isn't necessary, IMO it's fine to return the object itself.

As for the moment.js stuff, IIRC @brianhelba used a technique to jettison the locale stuff in another repo, but I can't remember which one. Or maybe I am mis-remembering this altogether...

matthewma7: Sounds good. I'll replace the lodash code. The reason @jbeezley went the extra mile to make a copy is probably to make sure nobody accidentally modifies stuff, but I agree it's ok to not worry about that too much. 
For momentjs locale bloat, there are kind of well-known existing solutions out there, I used the first one, it worked like that. https://github.com/jmblog/how-to-optimize-momentjs-with-webpack

zachmullen: > For momentjs locale bloat, there are kind of well-known existing solutions out there, I used the first one

Great! Seems straightforward enough

Looks like we still need to do a `yarn remove lodash` to clear out this change.

matthewma7: Yes, I am going to remove lodash stuff.

matthewma7:
Updated again. 

Please take a look.

PS: This is useful
https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore

PR #128, opened 2019-07-08
[WIP] Initial Access Control component

matthewma7:
This is an initial implementation of Access Control based on internal design, which is available on figma.com and on current girder client.

This PR
- Adds an initial Access Control component which has all functionality of its counterparts in existing girder client, except for the [custom access flags](https://girder.readthedocs.io/en/stable/plugin-development.html?highlight=custom%20access%20flag#adding-custom-access-flags), which, at the time of writing, I don't know if we are using and want to maintain in future. 
- Expose the Access Control component as an option of the DataBrowser component.
- Enhances the Search component to make it more flexible to use for this case and other future places. 
- Makes the DataBrowser show non-public item with the same fashion as the existing girder client, light yellow background color.
- Upgrades to latest Vue and Vuetify, as Vue 2.6 is stabilized and we should start to use new slot syntax, and Vuetify version has progressed a lot. 
- Includes some other small improvements that I gathered from my usage experience with GWC

This is a WIP PR please let me know for any feedback or question.


jeffbaumes:
This looks amazing.

Does the settings icon next to the grant access field do anything?

A tiny change I'd suggest is to make the search results placeholder dropdown shorter (maybe two entries?) so it's more subtle. Seems to make sense for it to get longer when the search request returns, as opposed to getting shorter.

matthewma7:
@jeffbaumes Thanks!
Allow me to use a GIF to explain the settings icon. 

The search component of the users and group is a reuse of the existing Search component (with some enhancement). 
The placeholder dropdown with grey placeholding items is a feature @subdavis and Jared specially made, which I really like, but as you now pointing this out, I think the placeholder is a little bit too long, not only for this case but also for its original use. So, I think I'll make update the Search component to make it shorter altogether (instead of providing a prop like palceholder-rows).

jeffbaumes:
That's great that you're reusing the search component from the main toolbar.

Since that placeholder is not specifically part of this branch, I'm good with that being a separate low-priority PR.

matthewma7:
> That's great that you're reusing the search component from the main toolbar.
> 
> Since that placeholder is not specifically part of this branch, I'm good with that being a separate low-priority PR.

Since I already changed the Search component within this PR, so I'll change it and updated the original GIF.

src/components/AccessControlDialog.vue
```diff
@@ -0,0 +1,291 @@
...
+      );
+      this.$emit('input', false);
+      this.$emit('model-access-changed', this.model);
+    },
+  },
+};
+</script>
+
+<template lang="pug">
+  v-dialog(
```
subdavis: So far I've avoided wrapping components in `v-dialog` and have put this burden on the consumer app.  I think it would be good to be consistent with `UpsertFolder` and  `Upload` and remove the dialog wrapper so that permissions could potentially be used in other ways.  I like the idea of this component taking only a `model`.

subdavis:
@matthewma7 wow.  This looks amazing.  Great work.

src/components/AccessControlDialog.vue
matthewma7: A good question! The reason I put it in a dialog and even suffix the naming of the component Dialog is that I can't think of a situation a consumer of this component won't use this in a dialog. I have used `UpsertFolder` and `Upload` as standalone components (not with DataBrowser) in other projects, but they all have been put in dialogs. I personally believe GWC should provide easy to use components and common sense convention, but If we would use this not in a dialog, I am happy to extract the component but would still keep the dialog component for ease of use.

subdavis: I've used Upload standalone without a dialog.  Also, dialog comes with a lot of configuration that users may want to change (persistent, width, scroll, etc.) and we don't want to restrict that or forward it through our own props, so IMO it's best to let apps set up their dialog.

matthewma7: I would say that hiding the ability to control these props enables me to provide easy to use common sense component. To be specific, for the width, I know the content won't look too good on a wide space so in this case, I decided for them (max-width 800px). For scrollable, because of the list of permission could go long, so I provided a scroll on the content. Also, for dialog scrollable to work, the content needs to have a v-card-text, which is hidden to the consumer unless open up the source code. 
But again, this is based on the assumption that the access control is better to be used in a dialog. 
The repo is not a general UI library, so I think should provide really easy to use components that simply works like that. 
This is the same reason I internalized `UpsertFolder` and `Upload` dialog to `DataBrowser`  https://github.com/girder/girder_web_components/pull/125. Because for the three times I use DataBrowser, I always wanted `UpsertFolder` and `Upload` to show just like they are in the demo app, but instead I needed to create the dialogs, hooking up the callbacks correctly, handling show state myself three times. By internalizing the dialog, the saved effort is huge. (See code reduction of demo/App.vue https://github.com/girder/girder_web_components/pull/125/files?file-filters%5B%5D=.vue)

PR #129, opened 2019-07-10
Move build-time packages to devDependencies

zachmullen:
It seems like these should be in devDependencies, so I'm assuming this was just an oversight.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,9 +20,7 @@
     "qs": "^6.5.2",
     "vue": "^2.5.21",
     "vue-async-computed": "^3.4.1",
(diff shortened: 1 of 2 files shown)


matthewma7:
I took a quick look of the test errors, I don't know if they are caused by @vue/test-util upgrade or Vuetify upgrade. @vue/test-util caused some problem in ResonantGeo, but Vuetify upgrade should be ok because I upgraded Vuetify in another PR.
Is it possible to manually move these two to dev without touching yarn.lock?

subdavis:
Looks like this has been done in #154

PR #133, opened 2019-07-22
Update boolean props

matthewma7:
This PR 
- Fix an issue related to breadcrumb. 
- Improved the girder-breadcrumb component. There always has been console warnings complaining about deprecated slot usage with v-breadcrumb. This commit fixes that issue and prepared for the new slot syntax. This commit also fixes a couple of issues introduced in the previous PR.
- Makes boolean props always negate its default value, so their default value can be deduced from the naming itself, and that when using the props, only the name is needed.
This commit also removes a couple of unused props. e.g. one of the props was called newFolderEnabled but it has a default value of true, so, it has to be set as :new-folder-enabled='false' to use it. It has been renamed to `new-folder`, to indicate it's default value is false, and when using, only need the mention the prop without value <girder-data-browser new-folder />

### DIFF
--- a/demo/App.vue
+++ b/demo/App.vue
@@ -17,7 +17,7 @@ v-app.app
               v-checkbox.mt-2(hide-details, label="Select", v-model="selectEnabled")
               v-checkbox.mt-1(hide-details, label="Draggable", v-model="dragEnabled")
               v-checkbox.mt-1(hide-details, label="New Folder", v-model="newFolderEnabled")
-              v-checkbox.mt-1(hide-details, label="Upload", v-model="newItemEnabled")
+              v-checkbox.mt-1(hide-details, label="Upload", v-model="uploadEnabled")
               v-checkbox.mt-1.mb-1(hide-details, label="Search Box", v-model="searchEnabled")
               v-divider
               v-list(dense)
@@ -54,10 +54,10 @@ v-app.app
           girder-data-browser(
               ref="girderBrowser",
               :location.sync="location",
-              :select-enabled="selectEnabled",
+              :selection="selectEnabled",
               :draggable="dragEnabled",
-              :new-item-enabled="newItemEnabled",
-              :new-folder-enabled="newFolderEnabled",
+              :upload="uploadEnabled",
+              :new-folder="newFolderEnabled",
               @click:newitem="uploader = true",
               @click:newfolder="newFolder = true",
               @selection-changed="selected = $event")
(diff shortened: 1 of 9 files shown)


matthewma7:
Hi @subdavis, this is ready for review.

src/components/DataBrowser.vue
```diff
@@ -324,21 +316,21 @@ girder-data-table.girder-file-browser(
...
               @crumbclick="changeLocation",
-              :no-root="noRoot")
+              :root-location="!noRootLocation")
           v-spacer
           slot(name="headerwidget")
           v-btn.ma-0(flat,
               small,
               color="secondary darken-2",
-              v-if="newFolderEnabled && nonRootLocation",
+              v-if="newFolder && isNonRootLocation",
```
subdavis: I'll remove these buttons completely in the next PR -- that's what we decided on right?

demo/App.vue
```diff
@@ -54,10 +54,10 @@ v-app.app
           girder-data-browser(
               ref="girderBrowser",
               :location.sync="location",
-              :select-enabled="selectEnabled",
+              :selection="selectEnabled",
```
subdavis: `selection` is a little ambiguous, easy to confuse with `selected`.  `selectable` would be preferred to match `draggable`, otherwise keep `select-enabled`.

tests/unit/DataBrowser.spec.js
```diff
@@ -69,11 +69,6 @@ describe('DataBrowser', () => {
           _modelType: 'user',
           _id: 'foo_user_id',
         },
-        selectEnabled: false,
-        multiSelectEnabled: false,
-        uploadEnabled: false,
-        newItemEnabled: false,
-        newFolderEnabled: false,
```
subdavis: I don't think it's a good idea to remove these completely.  Prefer explicitly setting these in the tests rather than relying on the default values.

src/components/DataBrowser.vue
matthewma7: Yes

demo/App.vue
matthewma7: `selectable` does sound better, I'll update it.

tests/unit/DataBrowser.spec.js
matthewma7: Hm, I'll add them back to make it more explicit.

src/components/DataBrowser.vue
matthewma7: But you will port them to the Concrete version, right?

subdavis: Yep!

matthewma7:
Thanks! I'll merge this now.

PR #135, opened 2019-07-25
README screenshot

subdavis:
I thought README.md needed a nice screenshot of GWC to add a little jazz.  I think a screenshot will make people more likely to try the demo or at least invoke a feeling of "hey this looks polished".

You can see how it looks [here](https://github.com/girder/girder_web_components/blob/026e72c573df87f9efc1111f430bb3ccc45f0903/README.md)


zachmullen:
Great idea! There has hitherto been zero marketing material in this repo, people would really have no idea what it's even useful for. I think we should also add two more screenshots to cover the "big three" components, with a little label under each one, "DataBrowser Component", "Authentication Component", "Upload Component".

subdavis:
Agreed.  I'll add the other 2 in some kind of configuration.

matthewma7:
This looks great!

But maybe we want to use a GitHub issue or this PR to hold the image, so the screenshot image won't be part of the source code?

subdavis:
@matthewma7 referencing an image by relative path is a neat feature supported by github to enable exactly this.  I don't see any reason not to keep documentation like this checked in.

matthewma7:
The reason is that it bloats up the repo and git history? When we update this image in the future, versions of the image will still live inside git history.

waxlamp:
`yarn.lock` by itself is 367K, while this screencap is only 33K. If we're talking about updating the image in every PR, then I can see how the bloat would become a problem but for now I don't see why we couldn't just include the image in the repo.

zachmullen:
Let's just put any smallish binary assets in a single directory, that would make it easy to prune from history later if we decide they're causing too much bloat.

subdavis:
Do this after #154, since things will look different.

subdavis:
Seems unnecessary given gwc.girder.org

It's not like we're trying to sell this externally.  GWC already has decent brand awareness, I don't think it would be worth the effort of keeping it updated.

PR #146, opened 2019-08-12
Serve from root

subdavis:
I've gone ahead and cut over the pages site to gwc.girder.org to trigger certificate generation on github's side.  I _think_ this is configured correctly, but we'll just have to wait and see.

Meanwhile, these changes are needed.


zachmullen:
Merged to test if this works

subdavis:
Looks like cert generation completed too.

PR #149, opened 2019-08-19
Token from oAuth reply

### LINKED ISSUE #73: Google OAuth login does not work on GH pages deployment
It looks like the whole oauth login flow goes through, but the token doesn't get set on redirect apparently.

subdavis:
Nothing too special here.  Worth noting that I think it's sufficient to do a rewrite on the `location.hash` of the page after getting the token from there.  This library isn't in control of page location, so there may be virtual hash routes from a vue router we don't want to change.

Here's a complex example:

```text
// before redirect
http://origin.com/path?query#/virtual/path#girderToken={girderToken}__
// after first redirect @ girder, girder fills the template string.  It may also append whatever other noise it wants -- that's the reason for the suffix.
redirect_to = http://origin.com/path?query#/virtual/path#girderToken=tokenstring__#other/stuff?
// after GWC has loaded, location should look like this:
http://origin.com/path?query#/virtual/path#other/stuff?
```

fixes #73

### DIFF
--- a/src/rest.js
+++ b/src/rest.js
@@ -20,7 +20,7 @@ function setCookieFromHash(location) {
   const token = arr[arr.length - 1].split(OauthTokenSuffix)[0];
   if (token.length === GirderTokenLength) {
     const expires = new Date();
-    expires.setDate((new Date()).getDate() + 7);
+    expires.setDate((new Date()).getDate() + 365);
     setCookieFromAuth({ token, expires });
     location.hash = location.hash.replace(`${OauthTokenPrefix}${token}${OauthTokenSuffix}`, '');
   }
--- a/src/rest.js
+++ b/src/rest.js
@@ -3,8 +3,28 @@ import cookies from 'js-cookie';
 import { stringify } from 'qs';
 import Vue from 'vue';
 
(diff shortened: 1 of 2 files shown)


src/rest.js
```diff
@@ -3,10 +3,30 @@ import cookies from 'js-cookie';
...
+/**
+ * set cookie if special string is found in the hash.
+ * @param {Location} location
+ */
+function setCookieFromHash(location) {
+  const arr = location.hash.split(OauthTokenPrefix);
+  const token = arr[arr.length - 1].split(OauthTokenSuffix)[0];
+  if (token.length === GirderTokenLength) {
+    const expires = new Date();
+    expires.setDate((new Date()).getDate() + 7);
```
subdavis: Don't have a way to get the expiration time of this token, so make it 7 days.  Could potentially add this as another format string to girder?

zachmullen: We should set it for a long time IMO; if it expires on the server, the presence of the cookie won't cause any problems, it'll just get rewritten once they log in again.

subdavis: Made it a nice round year.

PR #153, opened 2019-08-29
WIP: Vuetify 2.0.0 and Vue 2.6.10

subdavis:
@matthewma7 I've started on this, but it's going to be pretty painful.


demo/App.vue
```diff
@@ -10,8 +10,9 @@ v-app.app
             :close-on-content-click="false",
             content-class="girder-search-arrow-menu",
             v-model="uiOptionsMenu")
-          v-btn(icon, slot="activator")
-            v-icon.mdi-24px {{ $vuetify.icons.more }}
+          template(#activator="{ on }")
+            v-btn(icon, v-on="on")
+              v-icon.mdi-24px {{ $vuetify.icons.values.more }}
```
matthewma7: I would like to note one thing here. v-icon could take strings directly, it was a side effect that it worked with or without the {{ }}. In Vuetify 2, the designed usage is v-icon with string `$vuetify.icons.more`, but there is a lookup logic happening on the string which seems a little bit unnecessary to me personally.

.env
```diff
@@ -1 +1 @@
-VUE_APP_API_ROOT=http://localhost:8080/api/v1
+VUE_APP_API_ROOT=http://localhost:8095/api/v1
```
subdavis: revert

src/components/Authentication/Login.vue
```diff
@@ -15,33 +15,32 @@
...
           type="password",
           label="Password",
           :rules="nonEmptyRules",
-          prepend-icon="$vuetify.icons.lock")
+          prepend-icon="$vuetify.icons.values.lock")
       v-text-field(
           v-if="otpFormVisible || forceOtp",
           v-model="otp",
           type="text",
-          mask="######",
```
subdavis: this feature was dropped, reimplement with `otpRules`

PR #154, opened 2019-09-04
Vuetify 2.0 and Vue 2.6

subdavis:


.env
```diff
@@ -1 +1 @@
-VUE_APP_API_ROOT=http://localhost:8080/api/v1
+VUE_APP_API_ROOT=http://localhost:8095/api/v1
```
zachmullen: What's the reason for this change?

subdavis: I just didn't go to the trouble of creating `.env.local` like a reasonable person and it got swept in.

src/components/DataBrowser.vue
```diff
@@ -197,7 +197,7 @@ export default {
         parentType: location._modelType,
         parentId: location._id,
         limit: itemsPerPage >= 0 ? itemsPerPage : null,
-        offset: (page - 1) * itemsPerPage,
+        offset: (page - 1) * Math.abs(itemsPerPage),
```
matthewma7: Note: This handles when itemsPerPage is -1 (indicating no paging) and the result could become -0 which causes a problem in unit testing.

tests/unit/Authentication.spec.js
```diff
@@ -87,10 +87,10 @@ describe('Authentication', () => {
     expect(wrapper.vm.otpFormVisible).toBe(true);
     wrapper.setData({ otp: 'foobar' });
     await flushPromises();
-    expect(wrapper.vm.otp).toBe(''); // Test masking
+    expect(wrapper.vm.$refs.login.inputs[0].validate()).toBe(false);
```
matthewma7: Note: Behavior changed in Vuetify, invalid input won't be removed automatically.

src/components/DataBrowser.vue
subdavis: `-0` _kills_ me.  Good catch.

zachmullen: This is what happens when we only use IEEE 754 for all numeric types.

matthewma7:
Is there anything else needed for this PR? Or we are testing it by using it?

subdavis:
@matthewma7 I don't think there's anything left to do here.  However, the code is full of `v-layout` usage that I haven't converted to `v-col` and `v-row` yet.  

`v-layout` appears to have been deprecated silently but still works the same.   I'm fine with moving that work elsewhere.

zachmullen:
> because they work very differently.

You mean `v-row` and `v-column` are not meant to be a replacement for `v-layout(row)` and `v-layout(column)`? What's the appropriate migration path?

matthewma7:
`v-layout`, `v-flex` to `v-row`, `v-col` wasn't super straight forward to me when I was using v-layout vertically. I could give it a try within this PR. And by the way, remove any mentioning of stylus since vuetify 2 switched to scss.

subdavis:
I believe they're intended that way, but I don't really understand how to replicate some vertical layouts.  In particular, fill-height is just gone.  Margins are also now a part of the grid directives, which I find myself fighting against every time I use one.

https://vuetifyjs.com/en/getting-started/frequently-asked-questions#how-to-center-vertically

That's the only place in the documentation that `v-layout` is mentioned.  I think it might be a bug.

matthewma7:
> which I find myself fighting against every time I use one

Exactly!

src/components/MarkdownEditor.vue
```diff
@@ -63,11 +63,11 @@ export default {
     v-spacer
     v-toolbar-items
       v-btn(text, @click="activeTab = 0", :class="{ active: activeTab === 0 }")
-        v-layout(align-center, justify-content, column)
+        v-row.flex-column.align-center.justify-content(no-gutters)
```
subdavis: `v-row.flex-column` makes no sense to me. 

What's the resulting style?

```
display: flex
flex-direction: column
```

What does the row part do?

subdavis: It also kinda upsets me that all of those helper classes add a bunch of `!important` everywhere.

zachmullen: Agreed, `!important` is a code smell when it's in library code. Usually indicates that one is subverting the normal rules of CSS that control style precedence, but in Vuetify's case it may just be their way of dealing with a huge amount of complexity and combinatorics. It's an unenviable job...

matthewma7: I think ultimately all of these change comes down to, quote "_The grid has been rebuilt modeled after bootstrap. The existing grid still works and needs some slight modifications._". And since bootstrap doesn't have attributes, so stuff got downgraded into helper classes. 😅
But all the `!important` does look scary to me.

subdavis:
@matthewma7 @zachmullen 

Where are we here?  I'm not exactly _happy_ with the state of things (mostly referring to my own changes) but I think this is holding us all up.  I have additional features I need, and cherry picking from a few different branches is getting old.

I see two options:

1) merge it, push it to NPM, and deal with any fallout as it comes up.   I don't really have the time to address any major issues that would come up in a code review this week, so we would be merging more or less as-is
2) Start a new beta branch of some kind.  This will be a new major or minor version anyway, so we could name it 1.1.0.  Merge this to that branch and allow that branch to accept new features such as #155 and #156.  Once we feel confident that everything is O.K., maybe in a few weeks, merge to master.  

I feel better about option 2, but only on a visceral level.

matthewma7:
@subdavis 
BecauseVuetify 1 and 2 can't be used at the same time. This branch is the defacto master branch I am using nowadays. I didn't encounter major issues anymore. So, option would work for me.

subdavis:
Updated the readme to comply with the new vuetify setup.  Added some docs about a-la-carte setup (could @brianhelba take a quick look at that please?)

Once I get @matthewma7 and @zachmullen to sign off, we can merge to master.  Thanks especially to @matthewma7 for all your help on this enormous task.

matthewma7:
I upgrade the Vuetify to the latest but it broke some unit tests (later fixed). In summary, the unit test issue is caused by the choice of vue-test-util that tries to do synchronized rendering by default, which doesn't work so well anymore.

package.json
```diff
@@ -1,49 +1,52 @@
 {
   "name": "@girder/components",
-  "version": "1.0.0",
-  "license": "Apache-2.0",
-  "main": "dist/girder.umd.min.js",
+  "version": "1.1.0",
```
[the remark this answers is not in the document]
subdavis: Yeah sure.

README.md
```diff
@@ -137,33 +137,43 @@ See [the demo app](demo/App.vue) for a more comprehensive example.
...
+> **Note:** You must use the mdi icon pack. Icon packs cannot be mixed.
+
+Any custom configuration must be passed to Vuetify **after** Girder web components is
 installed to Vue:
+
 ```javascript
-// This order is important
-Vue.use(Vuetify, appVuetifyConfig);
+import Vuetify from 'vuetify/lib';
+// Order is important: your custom config should come after Girder
```
matthewma7: It looks like since the config of Vuetify is not feed through Vue.use(Vuetify, config). The order doesn't matter anymore.

README.md
```diff
@@ -159,14 +159,12 @@ const appVuetifyConfig = merge(girderVuetifyConfig, {
...
-installed to Vue:
-
 ```javascript
 import Vuetify from 'vuetify/lib';
-// Order is important: your custom config should come after Girder
-Vue.use(Girder);
-Vue.use(Vuetify, appVuetifyConfig);
+new Vue({
+  provide: GirderProvider,
+  vuetify: new Vuetify(appVuetifyConfig),
```
subdavis: good catch.

PR #155, opened 2019-09-06
Allow adding additional slots into the job table

subdavis:
This will be blocked on the vuetify upgrade, but I needed a way to add an additional cell to job rows.

Block on #154


subdavis:
Screenshot from 2019-09-06 12-56-47

subdavis:
@matthewma7 ptal

PR #156, opened 2019-09-11
Add prop to support condensed view during upload

zachmullen:
When uploading hundreds of files at a time, showing all of them
in the list causes a very tall element, and also slows down the
browser due to reactively repainting such a massive list on the
timescales at which updates are occurring.

This commit adds a new prop `maxShow` (integer) to the Upload and
FileUploadList components that limits the max number of files
shown at a time in the list. Additional files are truncated to
an element in the bottom that just shows their count.

This doesn't change the default behavior, which remains to show
the entire list.

Additionally, this fixes a bug where individual file progress
was not being rendered in the progress circle component, and also
condenses the height of each file element in the list to make
it more compact in all cases.

One downside of this approach is that the transitions are no
longer animated in the condensed mode, as I wasn't sure how
that could be achieved. I leave that as a future consideration.

### DIFF
--- a/src/components/Snippet/FileManager.vue
+++ b/src/components/Snippet/FileManager.vue
@@ -47,6 +47,10 @@ export default {
       type: Boolean,
       default: false,
     },
+    uploadMaxShow: {
+      type: Number,
+      default: 0,
+    },
     uploadMultiple: {
       type: Boolean,
       default: false,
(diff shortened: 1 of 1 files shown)


zachmullen:
@subdavis PTAL, as this is made against the vuetify upgrade branch.

matthewma7:
It looks like this new prop is simple to understand so I personally think we could have the FileManger pass it through.

subdavis:
I think I agree with @matthewma7 here: pass it through and give it the same default.

PR #159, opened 2019-09-25
Allow the consumer to inject widgets into each row of an upload file list.

subdavis:
Work in progress.

Block on #154

### DIFF
--- a/src/components/Upload.vue
+++ b/src/components/Upload.vue
@@ -19,7 +19,7 @@ v-card.fill-height(flat)
     v-col
       slot(name="dropzone")
         dropzone(v-if="!files.length",
-            @change="filesChanged",
+            @change="inputFilesChanged",
             :message="dropzoneMessage",
             :multiple="multiple",
             :accept="accept")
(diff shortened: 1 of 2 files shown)


src/components/Upload.vue
```diff
@@ -28,8 +28,8 @@ v-card.fill-height(flat)
       v-alert(:value="true", dark, type="error") {{ errorMessage }}
         v-btn(v-if="!uploading", dark, outlined, @click="start") Resume upload
 
-    slot(name="files")
-      file-upload-list(v-model="files", v-bind="{ currentIndex, maxShow }")
+    slot(name="files", v-bind="{ files, setFiles, currentIndex, maxShow }")
+      file-upload-list(@input="setFiles", v-bind="{ value: files, currentIndex, maxShow }")
```
subdavis: This is actually a separate bugfix.  If files are deleted such that index >= files.length, the index should be changed.

zachmullen:
Do you have a screenshot for posterity?

subdavis:
Examaple usage: 

Screenshot from 2019-09-27 11-19-45

zachmullen:
Thanks! It's great to see this library proving its mission in the wild :)

PR #160, opened 2019-09-25
Add option to use Girder-Authorization for auth.

subdavis:
I don't like adding this back, but I can't function without it.  As long as girder supports this auth mechanism, I feel like having the option to use it is within scope here.

### DIFF
--- a/src/rest.js
+++ b/src/rest.js
@@ -7,6 +7,11 @@ const GirderTokenLength = 64;
 export const OauthTokenPrefix = '#girderToken=';
 export const OauthTokenSuffix = '__';
 
+// Girder's custom headers
+const GirderToken = 'Girder-Token';
+const GirderOtp = 'Girder-OTP';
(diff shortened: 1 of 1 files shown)


subdavis:
@zachmullen ptal when you have a chance.

zachmullen:
> but I can't function without it

Why is that? In what context do you need to use one over the other?

subdavis:
I think I mentioned it a while back, but I'm working behind a proxy that filters out all `Authorization` headers.  Under normal circumstances I would say such a patch has no business being in this library, but girder luckily supports it.

zachmullen:
Hopefully we aren't subverting some security policy by passing basic auth via a different header...

subdavis:
As far as I can tell, the proxy is just configured improperly.  It should pass Authorization along when a request comes in that is already authorized with the proxy through another means, such as with a cookie.

PR #162, opened 2019-09-27
Allow upstream to specify upload phrase

subdavis:
A customer adamantly wants a different phrase for that button.  This seems like a low maintenance burden.

### DIFF
--- a/src/components/Upload.vue
+++ b/src/components/Upload.vue
@@ -14,7 +14,7 @@ v-card.fill-height(flat)
 
     v-card-actions(v-show="files.length && !errorMessage && !uploading")
       v-btn(text, @click="files = []") Clear all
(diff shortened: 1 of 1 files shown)


src/components/Upload.vue
```diff
@@ -74,6 +74,10 @@ export default {
       default: null,
       type: String,
     },
+    uploadPhrase: {
```
zachmullen: I don't have strong feelings, but maybe something like "startButtonText" would be more descriptive?

subdavis: That works for me.

danlamanna: +1 That also seems to be the format vuetify uses for internationalization phrases - https://vuetifyjs.com/en/customization/internationalization.

PR #167, opened 2019-10-03
npmjs deployment with circleci

subdavis:
I'm tired of not having this :P

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@girder/components",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "scripts": {
     "serve": "vue-cli-service serve demo/main.js",
     "build": "vue-cli-service build --target lib --name girder src/index.js",
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,6 +40,20 @@ jobs:
             - "eb:dc:cd:0a:f4:99:56:3f:79:f7:22:a2:6c:21:56:c7"
       - deploy:
           command: .circleci/pages-deploy.sh
+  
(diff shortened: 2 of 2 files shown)


subdavis:
I've tested all the individual parts of this and they seem to work.  

To publish to npm, you'd need to a) update package.json's version and merge to master and b) push a tag on that commit, then c) merge to master.

I'm pretty sure tags won't survive a squash-and-merge so we will have to probably dedicate PRs explicitly for release.  

Ready for review.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@girder/components",
-  "version": "2.0.0",
+  "version": "2.0.1",
```
subdavis: I've already released this by hand while testing token auth, so it needs to be merged.

subdavis:
This may not actually work.   For the publish job to run, the CI runner should be invoked by a tagged commit on master.  As far as I'm aware, that condition is impossible to satisfy.  Either squash-and-merge destroys your tag or the merge commit created by GitHub is the target of the CI job, which is 1 commit ahead of the commit you would have tagged.

zachmullen:
Are we ok to just tag master after the version change gets merged?

subdavis:
I honestly don't know.  I believed until now that simply pushing a tag wouldn't cause CI to run again for a commit that had already passed.

If that's wrong, then yes.

zachmullen:
IIRC pushing a tag *does* trigger CI, I'm pretty sure that's how I do it for Girder.

subdavis:
Then this is ready for merge and we can give it a try.

PR #169, opened 2019-10-07
Use best practices for app layout in demo app

zachmullen:
Since this demo app is highly likely to be used as a canonical example for downstream apps, we should make it adhere to vuetify best practices, which I admittedly am not an expert on. However, these two changes do seem to be in line with their recommendations.

### DIFF
--- a/demo/App.vue
+++ b/demo/App.vue
@@ -1,40 +1,39 @@
 <template lang="pug">
 v-app.app
+  v-app-bar(color="primary", dark, app)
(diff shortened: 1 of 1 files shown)


subdavis:
because of the whitespace change, its hard to see exactly what changed.  is this `v-app-bar` and `v-content`?

zachmullen:
Yes, sorry, I put it as notes in the commit but didn't promote that up to the PR text

PR #171, opened 2019-10-07
Prop in Auth and Login components to hide Forgot Password

zachmullen:
In some cases (e.g. pluggable auth on the backend), downstreams won't want a forgotten password flow to go through Girder at all. Or, like in my case, the app just isn't mature yet so there isn't a forgot password view.

### DIFF
--- a/src/components/Authentication/Authentication.vue
+++ b/src/components/Authentication/Authentication.vue
@@ -7,7 +7,7 @@ v-card.girder-authentication-component
   v-tabs-items(v-model="activeTab")
     v-tab-item(key="login-box")
       girder-login(:oauth-providers="oauthProviders",
-          v-bind="{ forceOtp, forgotPasswordUrl, forgotPasswordRoute, showForgotPassword }",
+          v-bind="{ forceOtp, forgotPasswordUrl, forgotPasswordRoute, hideForgotPassword }",
           @forgotpassword="$emit('forgotpassword')")
     v-tab-item(key="registration-box", v-if="register")
       girder-registration(:oauth-providers="oauthProviders")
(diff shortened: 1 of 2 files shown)


src/components/Authentication/Login.vue
```diff
@@ -38,10 +38,11 @@
             :loading="inProgress")
           v-icon(left) $vuetify.icons.login
           | {{ otpFormVisible ? 'Verify code' : 'Login' }}
-        v-spacer
-        v-btn(
-            text, color="primary", :to="forgotPasswordRoute", :href="forgotPasswordUrl",
-            @click="$emit('forgotpassword')") Forgot Password?
+        template(v-if="showForgotPassword")
```
subdavis: I didn't realize until now that you could use `template` in this way.

src/components/Authentication/Login.vue
```diff
@@ -90,6 +91,10 @@ export default {
       type: Array,
       default: () => [],
     },
+    showForgotPassword: {
+      type: Boolean,
+      default: true,
```
subdavis: Per #123 we sort of established a negative default for props.  Could change this to `hideForgotPassword` or simply change the default to false.

I know there was another discussion about having strict false for boolean props, but I don't remember where it went...

src/components/Authentication/Authentication.vue
```diff
@@ -52,6 +50,10 @@ export default {
       type: Boolean,
       default: false,
     },
+    showForgotPassword: {
+      type: Boolean,
+      default: true,
```
subdavis: Ditto.

src/components/Authentication/Login.vue
zachmullen: Sure thing

src/components/Authentication/Login.vue
zachmullen: https://vuejs.org/v2/guide/conditional.html#Conditional-Groups-with-v-if-on-lt-template-gt

It's a good practice to avoid adding an unnecessary extra element level to the actual DOM.

PR #173, opened 2019-10-07
Add search result slot with demo

### LINKED ISSUE #165: Alter search results view?
Should we allow the search results view to be altered? E.g., to show a breadcrumb or to calculate some new text value based on the search results (say a function of the Item's metadata or path), or even a thumbnail?

Should this be in the library or project specific?

subdavis:
closes #165 

Added an example usage to the demo app.  I can remove it if no value is added.

Screenshot from 2019-10-07 15-28-43

### DIFF
--- a/src/components/Search.vue
+++ b/src/components/Search.vue
@@ -1,6 +1,7 @@
 
 <script>
 import { DebounceCounter } from '../utils';
+import { usernameFormatter } from '../utils/mixins';
 
 export const SearchModeOptions = [
   {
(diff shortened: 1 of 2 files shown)


src/components/Search.vue
```diff
@@ -89,10 +89,11 @@ v-row.align-center.girder-searchbar(no-gutters)
           v-for="r in quickResults",
           @click="$emit('select', r)",
           :key="r._id")
-        v-list-item-action.mr-2
-          v-icon {{ $vuetify.icons.values[r._modelType] }}
-        v-list-item-content
-          v-list-item-title {{ r.name || formatUsername(r) }}
+        slot(name="searchresult", v-bind="r")
```
matthewma7: This feature looks good.
Could we rename the variable r to something like `record`?

PR #175, opened 2019-10-07
Flexible search mode and types

subdavis:
* Ability to optionally set the values and options for search mode and search type.
* Ability to enable and disable the menu.
* Lazy internal implementation.  Does not allow the consumer to "fix" or "disable" either form by keeping the value unchanged (same behavior as v-input).


src/components/Search.vue
```diff
@@ -36,12 +84,38 @@ export default {
...
+        this.$emit('update:searchMode', val);
+      },
+    },
+    internalSearchTypes: {
+      get() {
+        return this.lazySearchTypes;
+      },
+      set(val) {
+        this.lazySearchTypes = val;
+        this.$emit('update:lazySearchTypes', val);
```
zachmullen: Could we name this event just `update:searchTypes`? It seems that the internal nature of it is more salient than the lazy part.

subdavis: Done.

src/components/Search.vue
```diff
@@ -12,6 +44,22 @@ export default {
       type: String,
       default: null,
     },
+    searchModeOptions: {
+      type: Array,
+      default: () => SearchModeOptions,
+    },
+    searchMode: {
+      type: String,
+      default: '',
```
zachmullen: In other places throughout the code, I've used `default: null` to indicate a nullable field, should we do that here as well?

zachmullen: I'm going to amend this to say we should do that, after thinking about it more. `type: String` indicates that users of this component must *pass* a string if they intend to use this prop, but `default: null` provides a good out-of-band signal to indicate that no prop was passed at all. I think I like that convention.

PR #177, opened 2019-10-08
Publish to npm on package.json version update

### LINKED ISSUE #168: Broke CI
#68 broke CI because I fundamentally don't understand filtering.

Need to read https://circleci.com/docs/2.0/workflows/#executing-workflows-for-a-git-tag a few more times.

```yaml
            tags:
              only: /^v.*/
            branches:
              only: master
```
The above will run for master OR tag `v*`.  In order to run for just tags, I need to `ignore` wildcard branches, but it seems like that would run for tags matching the regex on any branch, so I (or anyone else) could push a tag to a feature branch and end up publishing to npm.

I'm considering writing or finding a utility to do the deployment only for the master branch on commit when the package.json version has upgraded compared to the latest version on the NPM repository.

This whole "manually push a tag that doesn't even have to match the the version you set in a possibly very old pull request" workflow is not what I want at all anyway. My goal is continuous deployment based on a democratic process like a pull requests.  Pushing tags is manual and unilateral -- I might as well be running `yarn publish` from the console.

Problems with the current approach:
* I could merge a PR that updates the version, merge 4 more PRs, then push the tag and cause publication of those 4 features.
* I could push a tag like v9.9.9 to branch `feature/foo` and cause publication I think?

As far as I can tell, looking at the ci config for girder, if I pushed a random tag version to a random branch of girder, it would get published to pypi.  It doesn't look like there's any requirement in the release workflow that the branch be master.

subdavis:
closes #168 

Need to verify that sed, awk, tr, grep, etc. are all installed on the node:10 image from circleci

### DIFF
--- a/.circleci/npmjs-publish.sh
+++ b/.circleci/npmjs-publish.sh
@@ -1,14 +1,14 @@
 #!/bin/bash
 set -e
 set -o pipefail
-export PACKAGE_VERSION=$(cat package.json \
+PACKAGE_VERSION=$(cat package.json \
   | grep version \
   | head -1 \
   | awk -F: '{ print $2 }' \
   | sed 's/[",]//g' \
   | tr -d ' ')
 
-export PUBLISHED_VERSION=$(npm show @girder/components version \
+PUBLISHED_VERSION=$(npm show @girder/components version \
   | tr -d ' ')
 
 if [ "$PACKAGE_VERSION" != "$PUBLISHED_VERSION" ]; then
--- a/.circleci/npmjs-publish.sh
+++ b/.circleci/npmjs-publish.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+export PACKAGE_VERSION=$(cat package.json \
+  | grep version \
+  | head -1 \
+  | awk -F: '{ print $2 }' \
+  | sed 's/[",]//g' \
+  | tr -d ' ')
+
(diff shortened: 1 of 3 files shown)


.circleci/npmjs-publish.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
```
danlamanna:
```suggestion
set -e
set -o pipefail
```

I think you want these at a minimum (possibly `set -u` as well), but I'm not sure if `set -e` will work because of https://github.com/koalaman/shellcheck/wiki/SC2155. ¯\_(ツ)_/¯

.circleci/npmjs-publish.sh
```diff
@@ -0,0 +1,21 @@
...
+set -o pipefail
+export PACKAGE_VERSION=$(cat package.json \
+  | grep version \
+  | head -1 \
+  | awk -F: '{ print $2 }' \
+  | sed 's/[",]//g' \
+  | tr -d ' ')
+
+export PUBLISHED_VERSION=$(npm show @girder/components version \
+  | tr -d ' ')
```
zachmullen: As advised in the blog @danlamanna linked, we ought to separate the `export` from the assignment, in case the shell would ignore failures from the right-hand command.

subdavis: Thanks I hadn't read that yet.  The export isn't needed at all, so I'll just drop it.

Issue #179, opened 2019-10-08
DataBrowser select behavior is strange


subdavis:
Currently the data browser allows selections to persist when you change working directories.  This behavior seems incorrect, but it has the benefit of being more flexible than the alternative.

For example, if DataBrowser were to reset the selection when the working directory changed, you'd need another prop to disable that behavior to get back to where we are now.  Currently, downstream can listen to location change events and reset the selected prop itself without any additional props.

Should we keep the current, wonky behavior for the sake of flexibility?

matthewma7:
Oh, yes. I personally had some trouble with the persisted selection exceptional when I need to do something with the selection in the custom app. I guess the best option is to make the persistent selection optional? and deselect by default to gain the flexibility of persistent selection and also don't contradict common expectations or add burden to the downstream consumers of the component.

zachmullen:
Given @matthewma7 's response, I'd say for now let's take no action, and wait until we hear from users if there is a problem.

subdavis:
Good enough for me :+1:

subdavis:
I was wrong, consumers have no control over the value of selected.  #181 addresses this.

matthewma7:
I don't know if I made myself clear. I had some difficulty with the current persisted selection behavior. The issue may be alleviated if there is some indicator to show that there are records selected.

subdavis:
> some indicator to show that there are records selected.

I think there was probably an issue with the data coming from $emit.update-selected being wrong.  Should be resolved by 181.  Is the ability to listen to selection change events sufficient for your needs?

matthewma7:
> > some indicator to show that there are records selected.
> 
> I think there was probably an issue with the data coming from $emit.update-selected being wrong. Should be resolved by 181. Is the ability to listen to selection change events sufficient for your needs?

We might be talking about different issues. I mean a user could select something in a directory then change to another directory. At that moment, the selection state is there but no visual indication of that.

subdavis:
>  there but no visual indication of that.

You'll see the indeterminate state of the checkbox in the header row, right?

matthewma7:
> You'll see the indeterminate state of the checkbox in the header row, right?

Hm, maybe that's a good point, but I kind of expect auto deselect or something more explicit about the selection as a new user, but once you know how it works, it could be ok.

jeffbaumes:
> You'll see the indeterminate state of the checkbox in the header row, right?

[...]

PR #181, opened 2019-10-08
Use v-model for data browser selected rows

### LINKED ISSUE #172: Use v-model for Data Browser selected array.
Allow for 2-way binding for what rows in the browser are selected.  Lets external operations like `delete` update the selected array.

subdavis:
resolves #172 #179

This is actually necessary to address needs in #179.  Without this change, it isn't possible for downstream to change the value of selected.

### DIFF
--- a/src/components/DataBrowser.vue
+++ b/src/components/DataBrowser.vue
@@ -44,6 +44,10 @@ export default {
       type: Boolean,
       default: false,
     },
+    value: {
+      type: Array,
+      default: () => [],
+    },
   },
 
   data() {
@@ -55,7 +59,7 @@ export default {
       internalRefreshCounter: 0,
       rows: [],
       rowsLoading: false,
-      selected: [],
+      lazyValue: this.value || [], // selected items
     };
   },
 
@@ -72,6 +76,15 @@ export default {
     internalSelectable() {
       return this.selectable && this.location && !isRootLocation(this.location);
     },
+    internalValue: {
+      get() {
+        return this.lazyValue;
+      },
+      set(val) {
+        this.lazyValue = val;
+        this.$emit('input', val);
+      },
+    },
   },
 
   asyncComputed: {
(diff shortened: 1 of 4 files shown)


zachmullen:
Heh, I literally just ran into this 5 minutes ago.

zachmullen:
To elaborate, a DANDI requirement is stated thus (emphasis mine):

> - when an item is selected from the search results list, the navigation should go to that items location in the hierarchy widget, **with that item's row highlighted**, and the details view for that item displayed

subdavis:
Turns out the select behavior was completely broken since Vuetify 2.

Vuetify now provides the header with methods to toggleAll.

This still isn't working perfectly, but there are a ton of tricky edge cases where v-checkbox can get "backwards" because its props don't match its internal state.  The checkbox can look "on" when its `value` or `inputValue` are false because DOM checkbox state still exists.

src/components/DataBrowser.vue
```diff
@@ -292,15 +296,15 @@ girder-data-table.girder-file-browser(
...
       th.pl-3.pr-0(width="1%", v-if="internalSelectable")
         v-checkbox.secondary--text.text--darken-1.pr-2(
             color="accent",
             hide-details,
-            :input-value="props.all",
-            :indeterminate="selected.length > 0 && !props.all",
-            @click.native="toggleAll")
+            :input-value="props.everyItem",
+            :indeterminate="internalValue.length > 0 && !props.everyItem",
+            @click.native="on['toggle-select-all'](!props.everyItem)")
```
subdavis: I'm confident I've done something real bad here, but I'm not sure what.

subdavis:
As discussed in #172 the default behavior is now to clear the selection when the working directory changes.  When moving between pages of results within the same directory, the vuetify-builtin behavior of DataTable will be used.

The ability to select items across folders is no longer supported.  If we need to enable this ability later, we can.

PR #183, opened 2019-10-09
Change from spread syntax to Object.assign

HastingsGreer:
We had significant issues with vue-template-es2015-compiler versioning, that came down to (unsuccessfully) allowing spread syntax to work in this file.
Could we just change it to an Object.assign? Thanks!

### DIFF
--- a/src/components/Presentation/FileUploadList.vue
+++ b/src/components/Presentation/FileUploadList.vue
@@ -10,7 +10,7 @@
           v-progress-circular(v-if="file.status === 'uploading'",
               color="primary",
               :rotate="-90",
-              :value="progressPercent({...file.progress, total: file.progress.size })",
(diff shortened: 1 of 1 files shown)


zachmullen:
We use the spread operator in many places in this library, so it will be difficult to keep future developers from adding it indiscriminately. Seems like it would be better to solve your downstream build issue, or as @subdavis has suggested in the past, ship a transpiled version of this library.

subdavis:
@HastingsGreer what version of vue-template-es2015-compiler are you using?

zachmullen:
Also if you post the error message you're seeing it could help diagnose your build configuration problem.

zachmullen:
I'm going to close this as rejection of this particular change, but feel free to continue to discuss the issue, or if there's some other fix we should do, please open an issue.

PR #189, opened 2019-10-11
possible fix

subdavis:
@zachmullen here's a possible fix that doesn't involve toggling the hash string.

It's based on the assumption that 
* your app being same-origin as girder is rare, so the default behavior should be to support cross-origin oauth
* Working out-of-the-box is better than feature toggle.

It simply swaps the order of how the token is obtained such that resetting the hash string is not conditional.

### DIFF
--- a/src/rest.js
+++ b/src/rest.js
@@ -22,12 +22,17 @@ function setCookieFromAuth(auth) {
  */
 function setCookieFromHash(location) {
   const arr = location.hash.split(OauthTokenPrefix);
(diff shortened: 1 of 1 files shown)


zachmullen:
This is a reasonable candidate solution, but I still don't feel good about us reaching in and messing around with `location.hash` as soon as a GirderRest client is instantiated. The GWC library doesn't assume a router or anything, so messing with `location.hash` at all feels dirty to me.

PR #193, opened 2019-10-18
Fix metadata alignment

subdavis:
# Before

Screenshot from 2019-10-18 15-03-30

# After

Screenshot from 2019-10-18 15-03-12

### DIFF
--- a/src/components/DataDetails.vue
+++ b/src/components/DataDetails.vue
@@ -20,7 +20,8 @@ v-card.data-details
     template(#row="props")
       v-row(justify="space-between")
         v-col.shrink.py-1.body-2.font-weight-bold {{ props.datum.key }}
-        v-col.shrink.py-1.body-2 {{ props.datum.value }}
+        v-col.py-1
+          v-row.mx-0.body-2(justify="end") {{ props.datum.value }}
   girder-detail-list(
       v-if="files.length",
       :title="`Files (${files.length})`",
--- a/src/components/DataDetails.vue
+++ b/src/components/DataDetails.vue
@@ -20,8 +20,7 @@ v-card.data-details
     template(#row="props")
       v-row(justify="space-between")
         v-col.shrink.py-1.body-2.font-weight-bold {{ props.datum.key }}
-        v-col.py-1
-          v-row.mx-0.body-2(justify="end") {{ props.datum.value }}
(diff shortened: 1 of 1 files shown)


src/components/DataDetails.vue
```diff
@@ -20,7 +20,8 @@ v-card.data-details
     template(#row="props")
       v-row(justify="space-between")
         v-col.shrink.py-1.body-2.font-weight-bold {{ props.datum.key }}
-        v-col.shrink.py-1.body-2 {{ props.datum.value }}
+        v-col.py-1
+          v-row.mx-0.body-2(justify="end") {{ props.datum.value }}
```
matthewma7: I definitely could have missed something here, but would simply remove the `.shirnk` works as well?

subdavis: Sort of, but that isn't exacly what I wanted.  In the "after" example above, `a` would be right next to `bvar` instead of right-aligned.  I wanted to have justify-content: space-between while still giving multi-line values room.

subdavis: re: adding text-right

That also isn't quite what I wanted.  The word "lazy" would end up right-aligned.

matthewma7: Hm, I see the problem of adding `.text-right` but still wondering if the wrapping element is necessary.

subdavis: Yeah I tried to avoid it too...  But I believe the wrapping element is necessary to produce exactly this layout.  You need the inner text box to grow to its natural width and then right-align the text container rather than the text itself.

There may be some CSS grid magic to avoid the extra node, but grid magic would be less readable than this.

matthewma7: Hm, I found a combination that works. `.py-1.body-2.d-flex.justify-end`

subdavis: > .d-flex

Brilliant!  Fixed.

matthewma7: Cool, learned something more about text-align with multiple lines. 😉

mgrauer:
Will a release get created from this or if not, what is the right
incantation in my downstream app’s package.json to use this off of the
master branch?

matthewma7:
@mgrauer I don't know about the release. 
but simply run `npm install git+https://github.com/girder/girder_web_components.git` to use the latest master always worked great for me.

PR #194, opened 2019-10-18
Add slot for no results in Search

subdavis:
Allow downstream to override the no-results row.

### DIFF
--- a/src/components/Search.vue
+++ b/src/components/Search.vue
@@ -177,7 +177,7 @@ v-row.align-center.girder-searchbar(no-gutters)
           v-list-item-content
             v-list-item-title {{ result.name || formatUsername(result) }}
       v-list-item(v-show="searchText && quickResults.length === 0 && !loading")
-        slot(name="noresults")
+        slot(name="noresult", v-bind="{ searchText }")
           v-list-item-action
             v-icon $vuetify.icons.alert
           v-list-item-content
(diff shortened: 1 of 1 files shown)


src/components/Search.vue
```diff
@@ -177,11 +177,12 @@ v-row.align-center.girder-searchbar(no-gutters)
           v-list-item-content
             v-list-item-title {{ result.name || formatUsername(result) }}
       v-list-item(v-show="searchText && quickResults.length === 0 && !loading")
-        v-list-item-action
-          v-icon $vuetify.icons.alert
-        v-list-item-content
-          v-list-item-title No results found for query '{{ searchText }}'
-          v-list-item-subtitle Modify search parameters or refine your query.
+        slot(name="noresults")
```
matthewma7: This simple change works just like that!
I am no expert, but I am wondering if it should be named `noresult`, singular instead of plural. 
And, maybe adding the searchText as a scoped slot variable would be more future proof?

subdavis: I chose `noresults` because that's what Vuetify calls it in their data tables: https://vuetifyjs.com/en/components/data-tables

I could change it to `no-results` to be completely consistent.

Agree with your `searchText` slot variable suggestion.

subdavis: scratch that, we've already got `searchresult` and `moreresylts` so there's no way to be consistent.  I'll change it to the singular.

subdavis: see 3239a25e38f96cb09ed374a593bfbc30f681bf9e

Issue #195, opened 2019-10-30
Support for Vuetify dark theme


matthewma7:
Vuetify has good built-in support for the dark theme. However, at this moment, GWC component doesn't support the dark theme well due to some hard-coded value or missing configuration.

matthewma7:
Specifically, we are working on an application that uses a dark theme. Most components work well with Vuetify dark theme but GWC does not.

![2019-10-30_15-56-36](https://user-images.githubusercontent.com/3123478/67894483-048e9a80-fb2f-11e9-82f9-18849d4b1100.png)

![2019-10-30_15-58-59](https://user-images.githubusercontent.com/3123478/67894548-21c36900-fb2f-11e9-951d-4c7f8f081d40.png)

![2019-10-30_15-59-35](https://user-images.githubusercontent.com/3123478/67894551-238d2c80-fb2f-11e9-921a-1892e62e6d54.png)

zachmullen:
Big +1 from me. Maybe we could make this a dynamic switch on our demo app so you can very easily see both modes. The goal should be simply setting `light` or `dark` on the very top `v-app` should be sufficient to make everything look correct. That might not be possible, but we should try and get as close to it as we can.

matthewma7:
Since the application decided to use a dark theme. If it's ok, I'll give it a try with the project today.

zachmullen:
One other thing -- if we manage to achieve this goal of supporting both modes via a top-level `v-app` switch, we should advertise that as a feature in the README.

matthewma7:
Hm, the font in the upload is inheriting its color so the contrast is a little low.
![image](https://user-images.githubusercontent.com/3123478/67949605-5af6ea00-fbbe-11e9-8606-9da99ec676e2.png)
Maybe, in this case, we could use a darker shade of grey when the dark theme is applied.

zachmullen:
If you want, I can make that change if you push your topic branch.

PR #197, opened 2019-10-31
Support Vuetify dark theme

matthewma7:
This resolves https://github.com/girder/girder_web_components/issues/195

At this moment, many custom color tunings are removed to make it more generic.
Some ternary conditions in the template and stylesheet are added to address some key differences.

### DIFF
--- a/demo/App.vue
+++ b/demo/App.vue
@@ -28,6 +28,8 @@ v-app.app
                     v-model="rootLocationDisabled")
                 v-divider.mt-2.mb-1
                 v-checkbox.mt-1.mb-1(hide-details, label="Search Box", v-model="searchEnabled")
+                v-checkbox.mt-1.mb-1(hide-details, label="Dark theme",
+                    @change="$vuetify.theme.dark=$event")
 
     v-spacer
     girder-search(v-if="searchEnabled", @select="handleSearchSelect")
--- a/demo/App.vue
+++ b/demo/App.vue
@@ -29,7 +29,7 @@ v-app.app
                 v-divider.mt-2.mb-1
                 v-checkbox.mt-1.mb-1(hide-details, label="Search Box", v-model="searchEnabled")
                 v-checkbox.mt-1.mb-1(hide-details, label="Dark theme",
-                    @change="$vuetify.theme.dark=$event")
+                    v-model="$vuetify.theme.dark")
 
     v-spacer
     girder-search(v-if="searchEnabled", @select="handleSearchSelect")
--- a/src/components/DataDetails.vue
+++ b/src/components/DataDetails.vue
@@ -30,7 +30,9 @@ v-card.data-details
       title="Actions",
       :clickable="true",
       :rows="actions",
-      :color="$vuetify.theme.themes.light.secondary.lighten5",
+      :color="$vuetify.theme.dark?\
+        $vuetify.theme.themes.dark.secondary.darken3:\
+        $vuetify.theme.themes.light.secondary.lighten5",
       @click="handleAction")
     template(#row="props")
       v-list-item-icon.mr-1
--- a/src/components/DataDetails.vue
+++ b/src/components/DataDetails.vue
@@ -30,9 +30,6 @@ v-card.data-details
       title="Actions",
       :clickable="true",
       :rows="actions",
-      :color="$vuetify.theme.dark?\
-        $vuetify.theme.themes.dark.secondary.darken3:\
-        $vuetify.theme.themes.light.secondary.lighten5",
       @click="handleAction")
     template(#row="props")
       v-list-item-icon.mr-1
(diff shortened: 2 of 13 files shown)


demo/App.vue
```diff
@@ -28,6 +28,8 @@ v-app.app
                     v-model="rootLocationDisabled")
                 v-divider.mt-2.mb-1
                 v-checkbox.mt-1.mb-1(hide-details, label="Search Box", v-model="searchEnabled")
+                v-checkbox.mt-1.mb-1(hide-details, label="Dark theme",
```
zachmullen: In my testing, `v-model="$vuetify.theme.dark"` worked.

src/components/DataDetails.vue
```diff
@@ -30,7 +30,9 @@ v-card.data-details
       title="Actions",
       :clickable="true",
       :rows="actions",
-      :color="$vuetify.theme.themes.light.secondary.lighten5",
+      :color="$vuetify.theme.dark?\
+        $vuetify.theme.themes.dark.secondary.darken3:\
+        $vuetify.theme.themes.light.secondary.lighten5",
```
zachmullen: We should make a semantic color for this using a custom key in each theme, see what I've done in 7bc217a9f4f162489b1cf2a7e6da358c76a88274

demo/App.vue
matthewma7: Good catch.

src/components/DataDetails.vue
matthewma7: Since we will have only dark and light two themes, would that be an overkill, and it may not work well in IE?

zachmullen: The theme shouldn't be referenced explicitly like this. Rather, it should just be `color="key"`, where key is some value that is present in each theme.

matthewma7: I thought using the theme color from the will make theming more easy to change. 
For example, If the user says that the secondary color is green, it will be a shade of green. Otherwise, the user needs to explicitly overwrite that key?

matthewma7: The user could use the theme generator and only pick the predefined colors. https://theme-generator.vuetifyjs.com/

zachmullen: @subdavis what do you think? I guess dark vs. light may need to be treated independently from dynamic dispatch of theme colors, but generally I would prefer to avoid these kinds of switch statements in components.

subdavis: `$vuetify.theme.currentTheme.secondary` seems to be what you want here.

matthewma7: @subdavis semantic colors? Could you explain a little bit more?

subdavis: right now we have primary, secondary, accent, etc.

But we do variations on those based on whether the theme is light or dark:

```js
$vuetify.theme.dark?\
        $vuetify.theme.themes.dark.secondary.darken3:\
        $vuetify.theme.themes.light.secondary.lighten5
```

@zachmullen has suggested and I think I agree that this should be avoided.

In most cases, we should be using `$vuetify.theme.currentTheme.<identifier>` so that we never have to add logic regarding light vs dark.  This solution doesn't help with the example above, though.

I think the best thing to do would be to add another named color in `vuetifyConfig.js` called `highlightedListItem` or something like that and specify an explicit value for it in each theme rather than trying to derive it from another color.

Another alternative is to drop the custom background styling on `DetailList.vue`.  It doesn't look as good (IMO) but fighting Vuetify never seems to result in clean code.

For now, I vote we drop the `color` prop and custom styling in `DetailList.vue`

matthewma7: @subdavis Cool, thank you for the clarification. 
We currently have another instance that needs this kind of theme checking. 
What do think about this case?
https://github.com/girder/girder_web_components/pull/197/files#diff-20a85c2a028ea24e3d92a12b008ceda7R302

matthewma7:
Hi @zachmullen, @subdavis had an offline discussion on the color conditionals and decided, as he suggested, to drop the color customization from the DataDetails component for now. The other conditional remaining is this one https://github.com/girder/girder_web_components/pull/197/files#diff-20a85c2a028ea24e3d92a12b008ceda7R302
Do you have suggestions on that one or for this PR in general?

zachmullen:
Regarding [this line](https://github.com/girder/girder_web_components/pull/197/files#diff-20a85c2a028ea24e3d92a12b008ceda7R302), I notice when inspecting the DOM that the `tr` tag created by that slot is not within the `thead` created by the underlying data table, it comes before it, which seems like a bug. I'm wondering if we get the header row inside the `thead` rather than outside of it, if the theming would look good automagically?

matthewma7:
@zachmullen. I see. I tried to remove the color all together from the header, but that seems to make the header too light.

matthewma7:
Thanks, @zachmullen, @subdavis for the review and inputs. 
I'll merge this now and let's what additional issue should be created.

PR #198, opened 2019-11-01
Dark theme support in Dropzone and switch in demo App

zachmullen:
Refs #195

### DIFF
--- a/src/components/Presentation/Dropzone.vue
+++ b/src/components/Presentation/Dropzone.vue
@@ -35,14 +35,13 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-$stripeColor: #f0f0f3;
 $img: linear-gradient(
   -45deg,
-  $stripeColor 25%,
+  rgba(160, 160, 160, 0.12) 25%,
   transparent 25%,
   transparent 50%,
-  $stripeColor 50%,
-  $stripeColor 75%,
+  rgba(160, 160, 160, 0.12) 50%,
+  rgba(160, 160, 160, 0.12) 75%,
   transparent 75%,
   transparent
 );
@@ -52,7 +51,7 @@ $img: linear-gradient(
   cursor: pointer;
   min-height: 260px;
   height: 100%;
-  background-color: #f6f6f9;
+  background-color: var(--v-dropzone-base);
   background-repeat: repeat;
   background-size: 30px 30px;
 
--- a/src/utils/vuetifyConfig.js
+++ b/src/utils/vuetifyConfig.js
@@ -9,6 +9,9 @@ import colors from 'vuetify/lib/util/colors';
 export default {
   theme: {
     dark: false,
+    options: {
+      customProperties: true,
+    },
     themes: {
       light: {
         primary: colors.lightBlue.darken1,
@@ -16,11 +19,13 @@ export default {
         accent: colors.lightBlue.darken1,
         error: colors.red,
         info: colors.lightBlue.lighten1,
+        dropzone: colors.grey.lighten3,
       },
       dark: {
         primary: colors.lightBlue.darken3,
         secondary: colors.grey,
         accent: colors.lightBlue.lighten1,
+        dropzone: colors.grey.darken2,
       },
     },
   },


demo/App.vue
```diff
@@ -27,7 +27,9 @@ v-app.app
                     label="Root Disabled",
                     v-model="rootLocationDisabled")
                 v-divider.mt-2.mb-1
-                v-checkbox.mt-1.mb-1(hide-details, label="Search Box", v-model="searchEnabled")
+                v-checkbox.my-1(hide-details, label="Search Box", v-model="searchEnabled")
+                v-divider.mt-2.mb-1
+                v-checkbox.my-1(hide-details, label="Dark", v-model="$vuetify.theme.dark")
```
subdavis: I had no idea you could v-model that.   Cool.

src/components/Presentation/Dropzone.vue
```diff
@@ -35,7 +35,7 @@ export default {
 </script>
 
 <style lang="scss" scoped>
-$stripeColor: #f0f0f3;
+$stripeColor: var(--v-dropzone-darken1);
```
subdavis: How does this work?

None of the modified dropzone colors are any of the colors listed at https://vuetifyjs.com/en/styles/colors so some mechanism is making up a whole new color palette (and not doing a good job, since lighten1 - lighten5 are all white).

I also think the new hover contrast is too high and doesn't look quite right, but I'm not sure how to deal with that other than adding another vuetify variable for "dropzoneHover".  I wonder if that may be the right thing to do, since whatever generated this palette produces step-sizes that are just too large to be useful to us. 

Screenshot from 2019-11-01 09-31-03


Screenshot from 2019-11-01 09-28-42

zachmullen: > How does this work? None of the modified dropzone colors are any of the colors listed

See my changes in `vuetifyConfig.js`. The colors I set as the `dropzone` value in both themes are already computationally modified from the color set.

> I also think the new hover contrast is too high and doesn't look quite right

I kind of felt the same way, but was trying to avoid having two different custom keys just for this one component. Maybe rather than using Vuetify's lighten and darken for the stripe color, we can do a SASS `darken` function to do one that is less of a dramatic step.

zachmullen: It also occurs to me that we are firmly in the realm of design here now that we're talking about theming and how many and what things should be expressed in a theme. I'd like to hear from @jtomeck on that specific issue.

subdavis: > See my changes in vuetifyConfig.js. The colors I set as the dropzone value in both themes are already computationally modified from the color set.

 I don't think I understand. If `$vuetify.colors.dropzone` is ` colors.grey.lighten3` AKA `#EEEEEE` AKA `--v-dropzone-base`, what is `--v-dropzone-darken1`?

It would intuitively be the same as `colors.grey.lighten2`, but it isn't.  It's actually just `hex(colors.grey.lighten3) - 0x282828` which isn't anything.

> we can do a SASS darken function to do one that is less of a dramatic step.

How?  `--v-dropzone-darken1` is defined at runtime, and the app theme is defined at runtime.   Even if you could import the light and dark variants at build-time, I'm not sure how you'd swap between with a css-only approach.

zachmullen: > It would intuitively be the same as colors.grey.lighten2, but it isn't.

I think the color transformations aren't perfectly uniform additive jumps in RGB space, they may occur in a different color space and change depending on a variety of parameters, with the intent of perceptual linearity.

> How? --v-dropzone-darken1 is defined at runtime, and the app theme is defined at runtime. Even if you could import the light and dark variants at build-time, I'm not sure how you'd swap between them with a css-only approach.

That's true, I forgot that this was using runtime CSS variables. It would need to be a CSS transform rather than a SASS transform, [but that does exist](https://topaxi.codes/modifying-css-colors-with-the-color-function/).

matthewma7:
Hi @zachmullen, I am wondering if you have some cycle to revisit this PR. 🙂 There is a plan to demo a downstream app that uses the dark theme with the upload control.

zachmullen:
To be honest I can't remember what needs to be done. What's holding this PR up? Feel free to take it over if you want it to be merged, I don't have lots of cycles before the end of this year.

matthewma7:
I am quite following the discussion above, but I am happy to take it over and try my best to implement the change.

subdavis:
@matthewma7 if you can fix the coloring re:above in a reasonable way, I'm good with it.  I think I'd endorse adding a new "named" color to vuetifyConfig.js for the stripe colors of the uploader.  It's low-tech and easy to understand, but feel free to try whatever.  Thanks!

matthewma7:
In the end, I handpicked a shade of grey for the strike that looks good in light and dark theme to me.

image

image

PR #199, opened 2019-11-13
bugfix remove height 100%

subdavis:
Not sure why that was ever there in the first place.  Breaks everything when the height of the container is explicit.


zachmullen:
I think this was necessary (or at least worked) to get the disappearing animation to work. I don't know if we even do those transitions anymore.

PR #203, opened 2019-11-19
Check permission before showing access dialog

matthewma7:
Fix the issue that the access control dialog shows on an item that doesn't permission

### DIFF
--- a/src/components/Snippet/FileManager.vue
+++ b/src/components/Snippet/FileManager.vue
@@ -99,6 +99,7 @@ export default {
       },
       actOnItem: null,
       showAccessControlDialog: false,
+      hasAccessPermission: false,
     };
   },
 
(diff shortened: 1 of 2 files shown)


src/components/Snippet/FileManager.vue
```diff
@@ -246,15 +247,18 @@ v-card.girder-data-browser-snippet
...
   v-dialog(
       v-model="showAccessControlDialog",
       max-width="700px",
       persistent,
+      eager,
       scrollable)
     girder-access-control(
+        v-if="actOnItem",
         :model="actOnItem",
+        :has-permission.sync="hasAccessPermission",
```
subdavis: Wont' this cause the dialog to flash while the request fires?  Like won't it open quickly then close again?

matthewma7: Hm, it is the disability of the list item that will change. It will flash from disabled to enabled but it's relatively mild.

matthewma7:
Thanks @subdavis 
I totally agree it feels odd to me to make a API requests for a context menu. 
I am hoping when we have more things in the context menu, we will be able to find a pattern for this sort of stuff. 🙂

PR #204, opened 2019-11-20
Add props to control browser page size

matthewma7:
This is a feature requested by a downstream project https://github.com/dandi/dandiarchive/issues/42
It looks like a good feature to have to GWC. However, am not sure if the effort and (technical debt) to implement reactive itemsPerPage can be justified, so the itemsPerPage prop only controls initial size. 

I'll be adding a test case for it, any comment or feedback is welcomed.

### DIFF
--- a/src/components/DataBrowser.vue
+++ b/src/components/DataBrowser.vue
@@ -48,12 +48,20 @@ export default {
       type: Array,
       default: () => [],
     },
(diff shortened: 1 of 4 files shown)


subdavis:
Approve of this as a concept.  I'll review code when you remove the [WIP] status.

Issue #205, opened 2019-11-25
Proper support of all page size in Data Browser


matthewma7:
It turns out Girder doesn't actually support 'all' option. (Is that true?). When no limit param is passed in, Girder will set it to 50. It looks like we should remove the 'all' option from GWC or add the 'all' option support to Girder.
@subdavis @zachmullen

subdavis:
Those params get pushed directly to the mongo query.  So yeah not providing the parameter would result in 50.   But you can send `0` as the limit and mongo will return the full list.

It's a leaky abstraction, but I suppose it works.

matthewma7:
Oh, 0 is interesting. I'll change the title and create a quick PR for it.

subdavis:
@matthewma7 I don't necessarily think that allowing a user to show "all" is a good idea. A max page size of 50 might be a better limit.  You could certainly lock up the page with too many elements for large folders.

Unless there's an explicit use case for seeing more than 50 items on a page, might be easier just to remove that from the list of options (as well as validate the parameter so parent components couldn't add it back)

matthewma7:
I think removing the 'all' from default makes sense, but if a downstream app chooses to do that, we should allow it?

subdavis:
I don't see how we could allow it without adding the `-1` to `0` transform in DataBrowser.vue, at which point we might as well support it fully?  I'd endorse either all or nothing.  Full support or no support.  

What do you think?

matthewma7:
I meant that we implement the `-1 to 0` magic, but don't provide the 'all' option by default, by changing the new [items-per-page-options](https://github.com/girder/girder_web_components/pull/204/files#diff-20a85c2a028ea24e3d92a12b008ceda7R55)'s default value.

subdavis:
That sounds fine.

PR #206, opened 2019-11-25
Fix issue that data browser don't handle page size 'all'

matthewma7:
While I was trying to add support for 'all' page size. 
I noticed that the browser makes unnecessary requests to folder or items under certain conditions, so I refactored the code to handle those conditions.

This PR also update default page sizes.

Resolve https://github.com/girder/girder_web_components/issues/205

### DIFF
--- a/src/components/DataBrowser.vue
+++ b/src/components/DataBrowser.vue
@@ -54,7 +54,7 @@ export default {
     },
     itemsPerPageOptions: {
       type: Array,
-      default: () => ([5, 10, 15, -1]),
+      default: () => ([10, 25, 50]),
     },
   },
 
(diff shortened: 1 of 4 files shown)


subdavis:
I'll need some time to review this.  The function you rewrote is, IMO, the most error-prone and annoyingly complicated code in GWC.

Could you give more details about the edge cases these changes address?

src/components/DataBrowser.vue
```diff
@@ -205,39 +205,37 @@ export default {
...
+        promises.push(this.girderRest.get(GIRDER_FOLDER_ENDPOINT, { params: folderParams }));
+      }
+      const itemOffset = folderOffset - nFolders > 0 ? folderOffset - nFolders : 0;
+      const numbersOfItemsWillBeFetched = nItems - itemOffset;
+      // if we could have items
+      if (location._modelType === 'folder'
+      // and there are items remaining
+      && numbersOfItemsWillBeFetched > 0
+      // and still have place for items
+      && (itemsPerPage === -1 || itemsPerPage - numberOfFoldersWillBeFetched > 0)) {
```
subdavis: This is an unusual style for multi-line conditionals.  Since airbnb is the Vue style guide we follow, https://github.com/airbnb/javascript/issues/1380 may be worth following.

https://github.com/airbnb/javascript#control-statements

matthewma7: Sounds good.

matthewma7:
Totally, this logic here is complex. 
The two cases were: 
When it's on a later page and the page doesn't contain folder anymore, a request for folder would still be made with an offset that is larger than 'nFolder' which means no record will be returned; When it's on a later page and there is no item in the folder at all, a request for item would still be made.

matthewma7:
Man, it really is complex. I force pushed some naming change trying to make it clearer.

tests/unit/DataBrowser.spec.js
```diff
@@ -336,12 +336,12 @@ describe('DataBrowser', () => {
           _id: 'fake_folder_id',
         },
         initialItemsPerPage: 20,
-        itemsPerPageOptions: [20, 30, 40, -1],
```
subdavis: Even though we aren't

matthewma7:
Thanks! This one is tricky to review.

PR #208, opened 2019-12-06
Color non public folder light yellow

matthewma7:
following the current convention, color non-public folder light yellow.

image

image

### DIFF
--- a/src/components/Presentation/DataTable.vue
+++ b/src/components/Presentation/DataTable.vue
@@ -145,6 +145,10 @@ v-data-table.girder-data-table(
       .nobreak {
         white-space: nowrap;
       }
+
+      &.itemRow.not-public {
+        background: rgba(255, 240, 156, 0.3);
+      }
     }
 
     td:first-child {


zachmullen:
Do we need to double-encode the private attribute? Seems we could do without a background color if we are OK with just using the icon to encode it, and that would simplify our theme palette.

zachmullen:
@jtomeck I'd love your opinion on my above question

jtomeck:
> @jtomeck I'd love your opinion on my above question

I'd definitely prefer if we just use the icon. If we did use a color as well I'd say using a shade darker grey would be better than the yellow.

matthewma7:
The yellow was simply because of existing girder implementation. 
The grey is being used for things like hovering state. 
Maybe we don't need this or do we have some other ideas? 🙂

zachmullen:
Having a hover color is great, let's remove the background color for private and just use the icon to encode that status.

matthewma7:
Totally, it's possible that the reason I am looking for the yellow background is simply that I am used to the existing conventions.

PR #209, opened 2019-12-20
Add an additional slot to DataTable

matthewma7:
This would allow the use of DataTable in a more customized way like this
image
Specifically, there is a downstream project that needs to distinguish folder navigation and opening the video clip within the folder.


subdavis:
This works for me, but could we name the slot something better?  Like `row-widget`?

I don't think additional is descriptive enough.

Implementation is fine.  Seems useful!

matthewma7:
@subdavis Thanks. row-widget sounds great. and we are using 'widget' throughout the codebase to indicate optional slot. 👍

Name changed.

PR #213, opened 2020-01-16
Fix bugs in demo app

zachmullen:


demo/main.js
```diff
@@ -10,8 +10,9 @@ const girderRest = new RestClient({
   apiRoot: process.env.VUE_APP_API_ROOT,
 });
 
-const notificationBus = new NotificationBus(girderRest);
-notificationBus.connect();
+const notificationBus = new NotificationBus(girderRest, {
+  useEventSource: true,
```
zachmullen: @subdavis this change isn't a bugfix, but I changed it while I was working in this file. Is there a good reason we don't use EventSource by default?

subdavis: I don't know.  I'm actually not familiar with the EventSource api.  I suppose the biggest reason we don't use it by default is Edge/IE compatibility.

Are we OK with that?

zachmullen: Oh, lame that Edge doesn't support it.

I guess most people use websockets, but I'm a big fan of EventSource since it just uses regular old HTTP requests, so no server-side support is required.

I'm fine either way; it's just a demo app so it doesn't really matter.

subdavis: /shrug let's make it the default.

PR #215, opened 2020-01-16
upgrade to vuetify 2.2.4

subdavis:
I feel like every successive update turns vuetify into a more and more quirky codebase, and I'm really starting to wonder if they've just lost control of the scope of the project and the organization is insufficient for the complexity they're managing.

I talked to Jared for a while and we are both really disappointed by the rigidity and high cognitive load of the v-row and v-col grid system compared to v-container and v-layout.  I'm about 1 free afternoon away from trying to replicate the 1.X v-container and v-layout components clean-room style in pure flexbox CSS.  Prepare yourselves for girder-container and girder-layout.

I also don't know how it makes sense that, if I specify a slot for the header, I also have to remember to disable the default header or else I get 2 headers.  That's not a sane default.

/rant

Upgrades to Vuetify 2.2.4

cc @HastingsGreer: I was not able to reproduce your error within the demo app with this upgrade.  This PR will, however, make your downgrade impossible.  So we'd have to get to the bottom of that issue.

EDIT: I'm actually able to reproduce it in the unit tests.  Fix hopefully incoming.

### DIFF
--- a/src/components/DataBrowser.vue
+++ b/src/components/DataBrowser.vue
@@ -324,21 +324,21 @@ girder-data-table.girder-file-browser(
     @dragstart="$emit('dragstart', $event)",
     @dragend="$emit('dragend', $event)",
     @drop="$emit('drop', $event)")
-
   template(#header="{ props, on }")
-    tr.secondary(:class="$vuetify.theme.dark?'darken-2':'lighten-5'")
-      th.pl-3.pr-0(width="1%", v-if="internalSelectable")
-        v-checkbox.pr-2(
-            color="accent",
-            hide-details,
-            :input-value="props.everyItem",
-            :indeterminate="internalValue.length > 0 && !props.everyItem",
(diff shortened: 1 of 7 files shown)


src/components/DataBrowser.vue
```diff
@@ -324,21 +324,21 @@ girder-data-table.girder-file-browser(
     @dragstart="$emit('dragstart', $event)",
     @dragend="$emit('dragend', $event)",
     @drop="$emit('drop', $event)")
-
   template(#header="{ props, on }")
-    tr.secondary(:class="$vuetify.theme.dark?'darken-2':'lighten-5'")
```
subdavis: This was just a collateral fix.  At some point in the past, this slot became responsible for adding the thead tag.  I don't think it makes a difference.

subdavis:
@zachmullen added the fix for Hasting's issue since that's the reason for this PR.

PR #219, opened 2020-01-23
Interactive documentation

subdavis:
This PR makes changes to the demo application.

I've found recently that when trying to explain girder web components, showing people the demo, which looks like a normal application, is confusing.  It's hard to explain to someone who isn't already a Vue developer what the "components" are and how they could be useful within your application.

I've also found that people are confused about what exactly this package offers.  Is it just a login box and the data browser? 

This PR seeks to make the demo app less of a "sample implementation" and more like interactive documentation. I'm really fond of what Vuetify has done, and that's the spirit of what I'm trying here.

Screenshot from 2020-01-23 15-32-43

I'm not done yet, and I didn't want to spend long on this (took about an hour), but hopefully this gives you an idea.

My end goal is:

1) a full enumeration of the top-level components with descriptions.
2) a couple of knobs for each one, showing some common toggles.
3) links to each component's code, so that the demo page can essentially be used as a reference when "I can't remember what props girder-upload takes"
4) include the npm/yarn install commands so the demo page looks more like a real library's landing page.  

Thoughts, @zachmullen and @jeffbaumes ?

### DIFF
--- a/demo/App.vue
+++ b/demo/App.vue
@@ -1,104 +1,176 @@
 <template lang="pug">
 v-app.app
-  v-app-bar(color="primary", dark, app)
-    v-toolbar-title Girder Web Components
-    v-toolbar-items
-      v-menu(
-          offset-y,
-          left,
-          :close-on-content-click="false",
-          content-class="girder-search-arrow-menu",
-          v-model="uiOptionsMenu")
-        template(#activator="{ on }")
-          v-btn(icon, v-on="on")
-            v-icon.mdi-24px $vuetify.icons.more
-        v-card
-          v-card-actions
-            v-row
-              v-col.py-0
-                v-checkbox.mt-1(hide-details, label="Worker Jobs", v-model="jobsEnabled")
-                v-divider.mt-2.mb-1
-                v-checkbox.mt-1(hide-details, label="Data Browser", v-model="browserEnabled")
-                v-checkbox.mt-1(hide-details, label="Select", v-model="selectable")
-                v-checkbox.mt-1(hide-details, label="Draggable", v-model="dragEnabled")
-                v-checkbox.mt-1(hide-details, label="New Folder", v-model="newFolderEnabled")
-                v-checkbox.mt-1(hide-details, label="Upload", v-model="uploadEnabled")
-                v-checkbox.mt-1(hide-details,
-                    label="Root Disabled",
-                    v-model="rootLocationDisabled")
-                v-divider.mt-2.mb-1
-                v-checkbox.mt-1.mb-1(hide-details, label="Search Box", v-model="searchEnabled")
-                v-checkbox.mt-1.mb-1(hide-details, label="Dark theme",
-                    v-model="$vuetify.theme.dark")
-
-    v-spacer
-    girder-search(v-if="searchEnabled", @select="handleSearchSelect")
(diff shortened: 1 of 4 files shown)


zachmullen:
Awesome, I really love this idea.

zachmullen:
Would be great if each component also had a link to its source (in the absence of having component API docs).

subdavis:
> Would be great if each component also had a link to its source (in the absence of having component API docs).

That's my intention.  Auto-generation of component API docs would currently be overkill, IMO.

subdavis:
Here's sort of where it ended up.  I'm happy with this as a first draft.

I also think it makes #135 a bit redundant now that the demo page has them all laid out.  Plus, we'd have to update the docs every time Vuetify's style updates, which is rather often.  

I actually don't want to duplicate information from the README into this page, because it will just be more stuff to have to remember to change, so I'm going to leave the npm information out.

Badges added for style and bandwagon-hopping, and for no real functional purpose other than signaling to other developers that we're cool.

Screenshot from 2020-01-23 19-35-40

danlamanna:
The badges really make it.

These docs look awesome! I would just mention that like all documentation, this adds burden. Someone has to be vigilant about ensuring that future PRs update documentation when appropriate.

subdavis:
> Someone has to be vigilant about ensuring that future PRs update documentation when appropriate.

I am willing to accept that.  I don't believe the maintenance cost of this is much higher than what we had before.  Prop changes will be the most common type of breakages for this page, and those are already exceedingly rare.

demo/App.vue
```diff
@@ -1,104 +1,181 @@
...
-        v-col(v-if="loggedOut", lg=4, md=5, sm=12)
+      v-col(xl=8, offset-xl=2, lg=10, offset-lg=1, md=12, offset-md=0)
+        .display-3 Girder Web Components
+        .title.mb-1 A Vue + Vuetify library
+          |  for interacting with
+          |
+          a(href="https://www.kitware.com/") Kitware's
+          |  data management platform,
+          |
+          a(href="https://girder.readthedocs.io/en/stable/") Girder
```
zachmullen: You may or may not already be aware of this, but there's an equivalent shorthand in pug for larger blocks of text:

```pug
        .title.mb-1.
          A Vue + Vuetify library for interacting with
          #[a(href="https://www.kitware.com/") Kitware's]
          data management platform,
          #[a(href="https://girder.readthedocs.io/en/stable/") Girder]
```

Not sure which is better, just wanted to make this feature known.

subdavis: I was not aware of this.   The thing I wrote was quite painful.  Thanks.

zachmullen:
This is looking awesome.

I think we should add a little bit of text at the top that indicates that these components are interacting with the live Girder instance running at data.kitware.com.

zachmullen:
And, we don't have to do it on this PR, but if we don't, let's add an issue for it: a ToC / Index navigation on the left would be great, we're already at a large enough number of components that it would help.

subdavis:
> I think we should add a little bit of text at the top that indicates that these components are interacting with the live Girder instance running at data.kitware.com.

:+1: 

Screenshot from 2020-01-24 15-09-22


> And, we don't have to do it on this PR, but if we don't, let's add an issue for it: a ToC / Index navigation on the left would be great, we're already at a large enough number of components that it would help.

I'd like to add that as time permits.  Prefer to make that a follow up.

Issue #161, opened 2019-09-26
Migrate templates from pug to html


zachmullen:
@subdavis @matthewma7 

I am a major fan of pug over HTML, but if I'm the only reason we are using pug in this project, I'd be fine to change to using HTML, which would slightly lower the burden of adopting this library from source downstream. How do you guys like it?

jbeezley:
I'm with you in disliking writing HTML tags, but I still don't like using Pug in Vue components for a few reasons:
1. Because it is a templating language, it conflicts with Vue's own templating language.  You can avoid using the templating features of Pug, but it can still be confusing to developers.
2. The standard tooling for linting, editor support, etc generally doesn't work well (or at all).
3. Downstream libraries not using Pug need to include extra webpack loaders.

matthewma7:
The pug did cause some issues regarding the new template syntax. I like the maximum compatibility of HTML in terms of dependency and ease of development when copy-pasting code snippets.

waxlamp:
We're using pug in Metabolomics, mostly on the strength of @subdavis's convictions, and with proper editor support I don't mind it. But:
- the linter support troubles are burdensome, to say the least
- removing a barrier to usage by "the general public" would probably be a good idea
- the linters for writing HTML in Vue components result in surprisingly readable code IMO (i.e., more readable than I might have assumed from the outset)

We're using HTML templates in Multinet and for some reason they just feel more solid to me there.

zachmullen:
Thanks for the opinions everyone. I certainly will not stand in the way of removing pug from GWC. It shouldn't be a breaking change to do so.

subdavis:
I also have a strong preference for pug as well for the sake of readability.

To respond to Jon's points: 

1. I have had no trouble distinguishing between the features of pug that I should and should not use, but I have seen that issue come up in other projects (metabulo).
2. I agree that pug's linting is terrible.  I haven't really felt any pain from this because I've internalized the syntax rules, but I can understand the barrier to entry that this may create.
3. I would be more concerned about having to include pug-loader in downstreams were it not for the fact that several other loaders are needed as well.  Elimination of pug and pug-loader from the dependency list wouldn't really impact the process of getting set up with with GWC.

I'm not convinced that GWC using pug provides any meaningful burden downstream, but I could be convinced to ditch it for the sake of our own developer experience as contributors to this library.

Also, a compiled es6 module library could solve a much larger amount of the adoption pain.

zachmullen:
[...]

PR #223, opened 2020-01-29
Add ToC to demo app

zachmullen:
This also contains a couple minor tweaks to the demo page, and
changes the job list component to say just "Jobs" instead of
"Girder Worker Jobs" since jobs are technically independent of
Girder Worker.

### DIFF
--- a/demo/App.vue
+++ b/demo/App.vue
@@ -35,7 +35,7 @@ v-app.app
               label="Dark theme",
               v-model="$vuetify.theme.dark")
 
-        a(id="auth")
+        a#auth
         headline(
             title="girder-auth",
             link="src/components/Authentication/Authentication.vue",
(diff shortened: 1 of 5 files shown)


zachmullen:
Closes #222 

I got the following form of pug lint failures at first:

```
38:11   Static attribute "id" must be written as ID literal
```

Right, I'm supposed to use `#` notation. The problem is, after doing that, I get:

```
38:10   ID literals must not be used
```

I understand that rule for components, but I don't see a way around it if we want to use anchors. My inclination would be to just disable this rule globally, but that feels heavy-handed. What do you think @subdavis ?

subdavis:
> My inclination would be to just disable this rule globally

Do it.  I'm starting to see merit in the argument against pug #161 

It was fine when I used to work entirely in pug.  Moving between pug and html is jarring now. I think whatever decision you make will be a temporary fix while we find the courage to re-write the templates in regular html.   Conversion should be pretty easy.

zachmullen:
Agreed, as time goes on it looks like pug is losing the battle for market share.

PR #224, opened 2020-02-07
Girder Jobs Status component refactor and JobFormatter mixin

subdavis:
Factor Girder Job object formatter into a mixin and the JobStatus into a component so I can show the status of a single job in consumers.

### DIFF
--- a/src/utils/mixins.js
+++ b/src/utils/mixins.js
@@ -1,5 +1,9 @@
+import moment from 'moment';
+
 import { AccessType } from '../constants';
+import * as jobstatus from '../components/Job/status';
 import Upload from './upload';
+
 /**
  * for components that need to show a locale-formatted date
  */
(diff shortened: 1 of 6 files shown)


zachmullen:
I'll review this in the morning, but it looks like the commit message got wonked maybe during a rebase.

subdavis:
@zachmullen ptal

src/utils/mixins.js
```diff
@@ -69,6 +73,34 @@ const usernameFormatter = {
...
+      return Object.assign({
+        statusText: statusDef.text,
+        statusColor: statusDef.color,
+        statusTextColor: statusDef.textColor || 'white',
+        statusIcon: statusDef.icon,
+        updateString: moment(job.updated).format('dddd, MMMM D, YYYY @ h:mm a'),
+        progressNumber: this.progressAsNumber(job.progress),
+        indeterminate: statusDef.indeterminate,
+        spin: statusDef.spin,
+      }, job);
```
zachmullen: I generally prefer `...` for multi-value assignment, but I don't feel strongly about it.

subdavis:
> making the icon rotate, but that can be done later and with help from the UI folks.

This is already supported, but only if the job moves through the "running" state.  If the job is too fast and moves from queued to success (or fetching input to success) then you don't see it.

src/utils/mixins.js
subdavis: :+1:

zachmullen:
Strange, when I manually set the job status to "2", the UI changes to reflect it (blue bar and the arrows icon) but no rotation is occurring. I do see the "rotate" class applied, but it's not rotating. Do you see that also?

PR #226, opened 2020-02-11
Results postupload

subdavis:
Adds results argument to postUpload


src/utils/mixins.js
```diff
@@ -236,7 +236,7 @@ const fileUploader = {
           }
         }
       }
-      await postUpload();
+      await postUpload(results);
```
zachmullen: To make this future proof let's do 
```javascript
await postUpload({ results });
```

PR #227, opened 2020-02-14
Bugfix/with credentials

subdavis:
client-side support needed to use changes from https://github.com/girder/girder/issues/2905

Edit: wrong reference.  https://github.com/girder/girder/pull/644

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@girder/components",
-  "version": "2.0.5",
+  "version": "2.0.6",
   "scripts": {
     "serve": "vue-cli-service serve demo/main.js",
     "build": "vue-cli-service build --target lib --name girder src/index.js",
--- a/src/rest.js
+++ b/src/rest.js
@@ -100,7 +100,7 @@ export default class RestClient extends Vue {
       headers[GirderOtp] = otp;
     }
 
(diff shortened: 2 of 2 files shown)


subdavis:
This fix isn't perfect.  It causes login to girder client as a side-effect, and if you log out of girder client, anything in your GWC app that relied on the cross-origin cookie will break until you log out and in again.

So this is really only half a fix.

zachmullen:
> It causes login to girder client as a side-effect

This seems like maybe a feature more than a bug, but either way it's OK behavior. Most users of Girder web apps aren't flipping back and forth between different client apps and logging in and out repeatedly.

subdavis:
@zachmullen something I didn't consider.  This will break GWC in all cases where Girder's CORS policy doesn't include an exact domain match.

For example, GWC will refuse to authenticate now for Access-Control-Allow-Origin = `*`.

That's definitely not what we want, so this should be optional and probably off by default.

zachmullen:
Good catch. Yes, let's make it optional. It probably makes sense on either the method or the class, but for practicality I'd vote to put the option on the class since that will make it so that downstream apps can configure it in their `main` method where the REST client is usually instantiated, which seems like where `process.env` references ought to be concentrated, rather than scattered throughout the code.

PR #228, opened 2020-02-14
Bugfix/with credentials optional

subdavis:
Follow up to #227, makes `withCredentials` optional and disabled by default.


src/rest.js
```diff
@@ -42,6 +42,9 @@ export default class RestClient extends Vue {
    *     (typically ending with "/api/v1").
    * @param {String} [opts.token] An initial value for the authentication token.
    * @param {Object} [opts.axios]  An axios instance to use for all requests.
+   * @param {Boolean} [opts.authenticateWithCredentials=false] Wehterh to set `withCredentials` on
```
zachmullen: *Whether

subdavis:
bump

PR #230, opened 2020-02-25
Parallel upload of files to girder

### LINKED ISSUE #231: Parallel file uploads
right now, the file uploader sends one at a time.  This is slow.

subdavis:
Still testing, but this is more or less done.

closes #231

### DIFF
--- a/src/utils/mixins.js
+++ b/src/utils/mixins.js
@@ -116,7 +116,7 @@ const fileUploader = {
       errorMessage: null,
       files: [],
       uploading: false,
-      currentIndex: 0,
+      indeterminate: false,
     };
   },
 
@@ -142,12 +142,18 @@ const fileUploader = {
   },
 
   methods: {
+    reset() {
+      this.files = [];
+      this.errorMessage = null;
+      this.uploading = false;
+      this.indeterminate = false;
+    },
+
     /**
      * Populate the internal list of files from an HTML FileList
      * @param {FileList} files from an input element
      */
     inputFilesChanged(files) {
-      this.currentIndex = 0;
       this.files = files.map(file => ({
         file,
         status: 'pending',
@@ -168,9 +174,6 @@ const fileUploader = {
      * @param {Array<Object>} files new file list
      */
     setFiles(files) {
-      if (this.currentIndex >= files.length) {
-        this.currentIndex = files.length - 1;
-      }
       this.files = files;
     },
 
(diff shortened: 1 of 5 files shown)


src/utils/mixins.js
```diff
@@ -183,60 +186,64 @@ const fileUploader = {
...
+              await file.upload.onError(error);
+              if (error.response) {
+                this.errorMessage = error.response.data.message || 'An error occurred during upload.';
+              } else {
+                this.errorMessage = 'Connection failed.';
+              }
+              file.status = 'error';
+              this.uploading = false;
+              this.$emit('error', { error, file });
+              throw error;
```
subdavis: This is behavior is different from what was here before.  Now, unless the caller of `start()` handles errors, this will be uncaught.   I tend to prefer that unhandled errors don't get swallowed by this library.

It will also be uncaught in Upload.vue.  I don't really see that as a problem, but we can catch it there if necessary.

src/utils/mixins.js
```diff
@@ -183,60 +186,64 @@ const fileUploader = {
...
-      const results = [];
+      const promiseList = [];
       this.uploading = true;
+      this.indeterminate = true;
       this.errorMessage = null;
       await preUpload();
-      for (; this.currentIndex < this.files.length; this.currentIndex += 1) {
-        const file = this.files[this.currentIndex];
+      this.indeterminate = false;
+      for (let i = 0; i < this.files.length; i += 1) {
```
zachmullen: This looks like it could be replaced with `const promiseList = this.files.map(...)`

mgrauer: I'm pretty sure `const promise list` is the name of a love song.

src/utils/mixins.js
zachmullen: I like the idea of throwing here, however, shouldn't we handle it in Upload.vue since that's where `start` is called? Otherwise wouldn't the only way to catch an exception during upload be to manually call `startUpload` from the outside?

subdavis: > shouldn't we handle it in Upload.vue since that's where start is called

Sure, an empty handler can be added to Upload.vue.

zachmullen: Sorry, I may be misunderstanding something. Prior to this change, my understanding was that the Upload.vue *did* handle the errors, by presenting the error message, and allowing "Resume" to be clicked. Was that not the case?

subdavis: No.  prior to this change, `start()` swallowed the error and expected the mixin consumer to present information about the error using `this.errorMessage`.  `errorMessage` is useful, but I also think `start()` should throw the exception too in case some external mixin consumer has other handling they want to do.

subdavis: I should say `start()` swallowed _some_ of the errors.  preUpload and postUpload were unhandled, but beforeUpload, afterUpload, and all axios errors were handled.

zachmullen:
Do we know why CI isn't reporting?

subdavis:
@zachmullen think it was just related to outages yesterday https://www.githubstatus.com/.  Webhooks were down, so CI might not have been reported.

I've squashed and added a release version.  Could you review again?  TY

PR #234, opened 2020-03-05
Hide the upload button if `startButtonText` is blank

waxlamp:
I've got a situation where I'd like to prevent the user from clicking on the stock "start upload" button, and I thought it would be neat if simply specifying a blank string as the button text would cause it not to be rendered at all.

### DIFF
--- a/src/components/Upload.vue
+++ b/src/components/Upload.vue
@@ -18,7 +18,7 @@ v-card.fill-height(flat)
 
     v-card-actions(v-show="files.length && !errorMessage && !uploading")
       v-btn(text, @click="reset") Clear all
-      v-btn(text, color="primary", @click="startUpload") {{ startButtonText }}
+      v-btn(v-if="startButtonText.length > 0", text, color="primary", @click="startUpload") {{ startButtonText }}
 
     v-col
       slot(name="dropzone")


subdavis:
At first it seemed like a nice idea, but I think an explicit toggle flag might be better.  In order to discover this feature, you have to read the button line and notice the v-if value that will cause the toggle.  The casual API reader is going to check the props to see what the component's capabilities are.  I also think having a prop will make accidental regressions less likely.

If this feels like overkill, I'd also be OK with a comment above the `startButtonText` prop as a message to future users that `Passing the value '' will cause the button to be hidden`.

waxlamp:
> At first it seemed like a nice idea, but I think an explicit toggle flag might be better. In order to discover this feature, you have to read the button line and notice the v-if value that will cause the toggle. The casual API reader is going to check the props to see what the component's capabilities are. I also think having a prop will make accidental regressions less likely.
> 
> If this feels like overkill, I'd also be OK with a comment above the `startButtonText` prop as a message to future users that `Passing the value '' will cause the button to be hidden`.

Doesn't feel like overkill; I agree that an explicit toggle would be better. I submitted this because it was "simpler" but I'll change it around to use a bool flag instead.

PR #235, opened 2020-03-06
release v2.1.0

waxlamp:


subdavis:
Why did you make this a minor and not a patch?

zachmullen:
We added a new feature, right? Semver would dictate a minor release for that.

PR #237, opened 2020-03-12
Remove unnecessary import

subdavis:
Because we're using vuetify-loader now, that import was unnecessary.  If a consumer wants to use this without vuetify-loader, importing the stylesheet will be their problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,18 +23,7 @@ yarn add -D sass sass-loader@^7.3.1 vue-cli-plugin-vuetify vuetify-loader
 ```
 
 > **Note:** If you are building with custom webpack (without vue-cli-service),
-> add the following to your module rules in webpack.config.js:
->
-> ```javascript
-> {
->   test: /\.(sass|scss)$/,
->   use: [
->       "style-loader",
->       "css-loader",
->       "sass-loader",
->   ],
-> },
-> ```
+> you should follow Vuetify's [Webpack install instructions](https://vuetifyjs.com/en/getting-started/quick-start/#webpack-install)
 
 ### Initialization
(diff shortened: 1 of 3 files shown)


zachmullen:
Do our README instructions for building need to change as a result of this?

waxlamp:
Just FYI, my immediate problem (which led to this PR as a test solution) is solved by simply rev-matching our project to the same version spec for the Vuetify dependency as the one GWC uses. That's clearly not a long-term solution, but that did work while the fix in this PR did not :crying_cat_face:

subdavis:
We don't explain anything about how to use this with webpack but without vuetify loader.  It wouldn't hurt to add, but this doesn't break docs.

subdavis:
@zachmullen updated readme to reflect these changes.  Note that our docs diverged from Vuetify's official docs, so I think it's best to just refer webpack users straight to vuetify.  That section includes the step of bringing in vuetify css.

zachmullen:
Isn't the sass-loader also necessary for our own components, since we use SASS?

subdavis:
Yeah, but since vuetify also needs sass and seems to have more particular needs, I figured it would be better to let their docs explain how to install it.

zachmullen:
Ok, we'll just need to keep our eyes open in case Vuetify decides to move away from SASS in the future.

subdavis:
I feel like the odds of them abandoning sass in anything other than a major version update (which would cause the majority of our docs would be invalidated) is really slim.

I'm also not aware of any GWC consumers who don't use vue cli (aside from maybe Forrest/Hastings that one time)

PR #238, opened 2020-03-18
Use HTML for templates instead of pug

### LINKED ISSUE #161: Migrate templates from pug to html

zachmullen:
Fixes #161

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,6 @@ jobs:
           paths:
             - project
       - run: yarn lint --no-fix
-      - run: yarn lint:pug
       - run: yarn lint:style
       # --runInBand https://discuss.circleci.com/t/babel-7-jest-results-in-timeout/25383
       # https://jestjs.io/docs/en/troubleshooting.html#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server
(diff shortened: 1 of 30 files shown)


zachmullen:
The test failures here are not a result of pug to HTML migration, but rather are a result of `yarn upgrade`. I haven't tracked down exactly which upgrade caused it. I suspected `vue-async-computed`, but downgrading that did not fix the failures.

zachmullen:
I backed out all the package upgrades, so this is now ready for review. I'm going to open a separate issue to address the upgrade issue I uncovered, but it shouldn't be encumbered by this large and unrelated PR.

One thing I'm unsatisfied with on this PR is that the auto-formatter I used output HTML tags with weird nesting, and then the linter's autofixing didn't adequately address the tag nesting, at least not in any way I find sensible. Furthermore, it expanded all the tags that didn't have a value part, e.g. rather than just say `small` it expanded all such attributes to `small="small"`, which is ugly and non-idiomatic for Vue.

I can try to address that on this PR by fiddling with the linter config, otherwise I can push it to another PR to isolate those changes.

PR #241, opened 2020-03-23
Autoformatting with js-beautify-html

subdavis:
Turns out autoformatting is hard.

* https://github.com/vuejs/vetur/issues/695 -- Vetur doesn't support a single autoformatter that is capable of running `eslint --fix` with our configuration.  You can try to find a prettier config that happens to overlap, but I don't believe it's possible.
* As a result, I've turned js autoformatting with vetur off for VS Code.  Eslint is much smarter than any prettier config.
* However, I did turn on lint-on-save features, which do read `eslintConfig` from package.json.  It's basically like running `yarn lint` every time you save

HTML formatting was another animal. Eslint isn't that opinionated about html formatting, and I think we should be more rigorous when possible.

For example:

```html
<v-tab-item key="login-box">
  <girder-login
            :oauth-providers="oauthProviders"
            v-bind="{ forceOtp, forgotPasswordUrl, forgotPasswordRoute, hideForgotPassword }"
            @forgotpassword="$emit('forgotpassword')" /> </v-tab-item>
```

eslint doesn't mind if you do stuff like this, but I'd prefer to be more strict:

```
      <v-tab-item key="login-box">
        <girder-login
          :oauth-providers="oauthProviders"
          v-bind="{ forceOtp, forgotPasswordUrl, forgotPasswordRoute, hideForgotPassword }"
          @forgotpassword="$emit('forgotpassword')"
        />
      </v-tab-item>
```

autoformatting with `js-beautify` will produce this, but more work is needed to figure out how to get it to run in CI.  I think this PR provides reasonable middle ground for VS Code users: hitting `ctrl+shift+i` will perform autoformatting.

I'd love to bring back `yarn lint:html` but I've spent too long on this this morning so far.  Settings for https://github.com/beautify-web/js-beautify can probably be found so that we can run `js-beautify` from a package.json command.

### DIFF
--- a/src/components/AccessControl.vue
+++ b/src/components/AccessControl.vue
@@ -155,7 +155,8 @@ export default {
         <breadcrumb
           :location="model"
           readonly="readonly"
-          no-root="no-root"/>
+          no-root="no-root"
+        />
       </div>
     </v-card-title>
     <v-card-text class="pt-0">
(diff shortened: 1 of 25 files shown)


src/components/AccessControl.vue
```diff
@@ -155,7 +155,8 @@ export default {
         <breadcrumb
           :location="model"
           readonly="readonly"
-          no-root="no-root"/>
+          no-root="no-root"
```
zachmullen: I was hoping there was some rule that would change these boolean properties back to just the `no-root` form. This is technically equivalent, just seems less preferable than the shorthand.

subdavis: Created #242 because I don't want to leave it like this, but I don't feel like fixing it right now.

PR #243, opened 2020-03-23
Include _id, _modelType in drop event payload.

subdavis:
replaces #120


cjh1:
Thanks @subdavis

subdavis:
@cjh1 it's not released yet.  I'll let you know when it is.

PR #247, opened 2020-03-25
Upgrade axios-mock-adapter and babel core

zachmullen:
Depends on #246


zachmullen:
This is ready for review. I tried my hand at upgrade `@vue/test-utils` as well, however it was getting more involved than I have time to deal with right now. Maybe someone braver than me can take that on.

PR #248, opened 2020-03-31
Support additional arguments for UploadCls

subdavis:
At the behest of @BryonLewis

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@girder/components",
-  "version": "2.2.0",
+  "version": "2.2.1",
   "scripts": {
     "serve": "vue-cli-service serve demo/main.js",
     "build": "vue-cli-service build --target lib --name girder src/index.js",
--- a/src/utils/mixins.js
+++ b/src/utils/mixins.js
@@ -185,6 +185,7 @@ const fileUploader = {
       preUpload = async () => {},
       postUpload = async () => {},
       uploadCls = Upload,
+      uploadClsArgs = {},
     }) {
       this.uploading = true;
       this.indeterminate = true;
@@ -210,6 +211,7 @@ const fileUploader = {
             $rest: this.girderRest,
             parent: dest,
             progress,
+            ...uploadClsArgs,
           });
           promiseChain = promiseChain
             .then(() => file.upload.beforeUpload())
--- a/src/utils/mixins.js
+++ b/src/utils/mixins.js
@@ -164,6 +164,7 @@ const fileUploader = {
         },
         upload: null,
         result: null,
+        uploadClsParams: {},
       }));
     },
 
@@ -185,7 +186,6 @@ const fileUploader = {
       preUpload = async () => {},
       postUpload = async () => {},
       uploadCls = Upload,
-      uploadClsArgs = {},
     }) {
       this.uploading = true;
       this.indeterminate = true;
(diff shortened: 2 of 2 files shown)


zachmullen:
I'm curious about the use case. If the arguments are known statically, we don't need this. If they're dynamic but invariant across all the files, then this makes sense.

@BryonLewis could you explain what you are trying to pass in your case?

subdavis:
@zachmullen is right.  @BryonLewis wants to be able to rename files when they're uploaded, so I wanted to provide a way to override the way upload classes get those things.

I've updated this PR with something that would actually work.

zachmullen:
Couldn't this be handled by lightly extending the Upload class? Or am I missing something?

BryonLewis:
> @BryonLewis could you explain what you are trying to pass in your case?

I'm utilizing the mixin to support parallel uploads in VIAME-Web.  We need the ability to change the upload name for the file.  By default the mixin relies on the File Object passed to it.  It's name property is read only.  I can think of two options to achieve this while still using the mixin:
I can create a new file object and copy over everything except the name.  
There might be a way to override the default Upload constructor used so I can pass a variable for the name.  

I don't believe either of those solve a problem if I want to append, prepend or modify all files in a list. (At least in a clean way)

Providing the option for the additional args seemed like the cleanest way to me.  I'm open to any other ideas for a solution.

subdavis:
> Couldn't this be handled by lightly extending the Upload class? 

I don't think so.  The `uploadCls` is constructed by the upload mixin, so even if you wrote an override of the `start()` method, you still need to attach some custom bit of data on the file that the desired new name can be stored in.

zachmullen:
If the file name being read-only is the thing that needs to be worked around, I think we could still do the subclass approach, by shallow copying the file in a way that should preserve its *behavior*, though not its type. If we're comfortable with that idea, the extension would just be:

```javascript
async start() {
  this.file = {
    ...this.file,
    name: <renaming logic here>,
  };
  return await super.start();
}
```

What do you guys think?

zachmullen:
Actually I just realized the spread operator doesn't work to shallow copy the File API, but if there's a relatively elegant way to do it, I'd still favor this approach.

subdavis:
> What do you guys think?

```js
async start() {
  this.file = {
    ...this.file,
    // this is the part I don't know how to make work.
    // all we have to work with is the file object, which may not be enough to determine the new name.
    // for example, the new name may be based on state in vuex or another component's data.
    // so we still need somewhere to put this on the file object from upload mixin.
    name: <renaming logic here>,
  };
  return await super.start();
}
```

zachmullen:
> I don't believe either of those solve a problem if I want to append, prepend or modify all files in a list. (At least in a clean way)

This is correct, to do something like that you'd hook into `preUpload` and `postUpload` hooks of the upload component itself.

BryonLewis:
`newFile = new File([oldFile], newFilename, {type: oldFile.type, lastModified:oldFile.lastModified});`

That is essentially what I was doing to the file**S** object before passing it to the mixin for uploading.  I don't know if I would call this elegant though. I just figured that if the default mixin's Upload class already had the ability to specify a name that it might be easier to expose that instead of doing this.

zachmullen:
Ok. If you think that's the easiest way I'm fine with this change.

PR #250, opened 2020-04-01
Expose job table as slot differently

subdavis:
* removes jobsWidget slot, since it was underpowered
* exposes jobs table as slot
* exposes jobs row as slot.

In order to use the jobs row slot, you have to also use the jobs table slot.  It may be reasonable to say that the jobs table doesn't provide enough value on its own to expose this, so if a consumer wants to extend the jobs table, they're on their own to just implement the whole thing.

### DIFF
--- a/src/components/Job/JobTable.vue
+++ b/src/components/Job/JobTable.vue
@@ -2,6 +2,20 @@
 import { jobFormatter } from '../../utils/mixins';
 import JobProgress from './JobProgress.vue';
 
+export const defaultHeaders = [{
+  text: 'Job Title',
+  value: 'title',
+}, {
+  text: 'Type',
+  value: 'type',
+}, {
+  text: 'Last Updated',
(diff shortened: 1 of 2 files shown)


src/components/Job/JobTable.vue
```diff
@@ -65,38 +59,37 @@ export default {
...
-          <slot
-            v-bind="props"
-            name="jobwidget"
-          />
-        </td>
-      </tr>
+      <slot
+        v-bind="props"
+        name="jobrow"
+      >
```
subdavis, line 73: I'm thinking this and the headers prop may be overkill, and if you want this table to look different, just write your own from scratch.

zachmullen: Do we have use cases in the wild for extending this component? I'd avoid speculating about it until we have at least one.

subdavis:
This is premature.

PR #252, opened 2020-04-27
Make download actions links

### LINKED ISSUE #225: Allow for copy link address in item details view, from download
You can see this behavior in gui.dandiarchive.org. If I go into an item details view, there is a download button, but I cannot get the link address target from the button (right clicking doesn't allow me the "copy link address" action). Could we enable the download link to be visible somehow (right click menu is fine)?

dchiquit:
In order to make the `View Item`, `Download`, and `Download (zip)`
actions more useable, embed their text into a link rather than
exclusively rely on using click events. This allows users to open them
in a new tab or copy the link address.

Functionally, the only change is that the action text is now rendered as a link, and behaves like a link when right-clicked. The rest of the action button reacts and behaves the same. I think it's useful for the text to be rendered as a link to indicate that the normal link actions can be done on it.

Screenshots taken from the demo app:
Screenshot from 2020-04-27 12-31-50
Screenshot from 2020-04-27 12-32-03

Closes #225
Also closes https://github.com/dandi/dandiarchive/issues/82

### DIFF
--- a/src/components/DataDetails.vue
+++ b/src/components/DataDetails.vue
@@ -82,7 +82,16 @@
           </v-icon>
         </v-list-item-icon>
         <v-list-item-content :class="`${props.datum.color}--text`">
-          {{ props.datum.name }}
+          <a
+            v-if="props.datum.url"
+            :href="urlForAction(props.datum)"
+            @click.prevent
+          >
+            {{ props.datum.name }}
+          </a>
+          <span v-else>
+            {{ props.datum.name }}
+          </span>
         </v-list-item-content>
       </template>
     </girder-detail-list>
@@ -95,12 +104,15 @@ import GirderMarkdown from './Markdown.vue';
 import GirderUpsertFolder from './UpsertFolder.vue';
 import { dateFormatter, sizeFormatter, usernameFormatter } from '../utils/mixins';
 
-function download(baseurl, modelType, id, query = '') {
+function generateUrl(baseurl, modelType, id, query = '') {
   if (['resource', 'folder', 'item', 'file'].indexOf(modelType) < 0) {
     throw new Error(`${modelType} is not downloadable`);
   }
   const idpart = id ? `${id}/` : '';
-  const url = `${baseurl}/${modelType}/${idpart}download${query}`;
+  return `${baseurl}/${modelType}/${idpart}download${query}`;
+}
+
+function download(url) {
   window.open(url, '_blank');
 }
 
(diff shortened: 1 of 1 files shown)


src/components/DataDetails.vue
```diff
@@ -148,43 +160,48 @@ export const DefaultActionKeys = [
     name: 'View Item',
     iconKey: 'view',
     color: 'primary',
-    handler() {
-      const { value: items } = this;
-      download(this.girderRest.apiRoot, items[0]._modelType, items[0]._id, '?contentDisposition=inline');
+    url(apiRoot, items) {
+      return generateUrl(apiRoot, items[0]._modelType, items[0]._id, '?contentDisposition=inline');
+    },
+    handler(girderRest, items) {
```
zachmullen, line 166: It seems like we can delete these `handler`s now, or am I missing something? Like we should make the `handler` optional, and change the docstring on L149 to reflect the new behavior. Then we could remove the `@click.prevent`, which seems like a good idea. At best, we want to `.stop`, not `.prevent` on something with an href.

dchiquit: `handler` still has a distinct implementation in the `Delete` action, so I left it in place. An alternative would be to change `handleAction` to test if there is a `url`, and if so, do `download(action.url(girderRest.apiRoot, items));` directly there. Then we could delete the stupid-looking `hander`s that I left hanging around. I didn't do that because I thought the link between having a `url` and being downloadable wouldn't be obvious.

I missed the docstring, I will update it to reflect the new schema.

I'm pretty sure that as long as we still have `handler`, the event modifier should still be `.prevent`. `.stop` will stop the event from bubbling up to the girder-detail-list.

zachmullen: I'm just suggesting that we make `handler` optional, such that these actions would either implement a `handler` or a `url`, but not both. Then inside `handleAction`, we only call the `handler` if one is set. Does that make sense? If we do that, we don't need `.stop` or `.prevent` at all.

> I'm pretty sure that as long as we still have handler, the event modifier should still be .prevent. .stop will stop the event from bubbling up to the girder-detail-list.

In that case, stopping the event bubbling up is exactly what we want, since it's the ancestor element that actually handles the event. But again, I think the right solution is to use neither and let the event process as normal.

subdavis:
I have meetings for the next few hours, but I want to look at this.  In particular, I wonder if it's possible to use the `href` prop on `v-list-item` to get the same effect and use vuetify tools rather than our own impl.

zachmullen:
+1 for using `href` on `v-list-item`

dchiquit:
I think that's the right solution. I will submit another PR for that fix.

dchiquit:
I set up a new branch and a new PR with a different approach: https://github.com/girder/girder_web_components/pull/254
closing this PR

PR #253, opened 2020-04-27
225 suggestion

subdavis:
@dchiquit I don't typically like to butt in on other people's features like this, but I needed to mess around with code to see if my idea would even work.

Here's a suggestion that I think simplifies things a bit.  The one part of your PR that I was opposed to was the use of `this` inside the `ActionKeys` handler function, because it isn't clear how the function context gets set and I found it confusing.  I know I did it that way initially, and that wasn't great.

Now, the `actions` computed prop adds a computed `url` to the action iff a `urlfunc` exists, and if that is found, the `v-list-item` `href` (and optionally, `target`) are used.

Let me know what you think.  We don't have to actually merge this PR, just thought it'd be helpful to see.

### DIFF
--- a/src/components/DataDetails.vue
+++ b/src/components/DataDetails.vue
@@ -82,16 +82,7 @@
           </v-icon>
         </v-list-item-icon>
         <v-list-item-content :class="`${props.datum.color}--text`">
-          <a
-            v-if="props.datum.url"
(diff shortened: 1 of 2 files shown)


dchiquit:
Closing in favor of https://github.com/girder/girder_web_components/pull/254

PR #255, opened 2020-04-28
[docs] document how to use girder media in page elements

subdavis:


subdavis:
It's possible that this section should go above the `For Developers` subheading.

README.md
```diff
@@ -211,6 +211,14 @@ the Girder API server running).
 When running against your own instance of the Girder API server,
 make sure to set [CORS](https://girder.readthedocs.io/en/stable/security.html#cors-cross-origin-resource-sharing) accordingly.
 
+### Displaying authenticated media from Girder on a page
+
+If your app injects media dynamically into the page using `img` or `video` elements that load from Girder, there are several requirements.
+
+* You must use the `authenticateWithCredentials` option in the GirderRest constructor.
```
zachmullen, line 218:
```suggestion
* If the data is private, you must use the `authenticateWithCredentials` option in the GirderRest constructor.
```

subdavis: This subsection is called "Displaying authenticated media from Girder on a page".  I believe if the data is public, you don't have to take any of these steps, is that right?

Should I rename the subsection "Displaying private media from Girder on a page" for clarity or "Displaying media from Girder on a page" and explain that, in the public case, everything should just work?

zachmullen: > if the data is public, you don't have to take any of these steps, is that right

If the data is public, you still need at least a wildcard ACAO, but you're correct that you don't have to take any of these specific steps, and that's out of scope for this section. Let's go with "displaying private media" and leave the instructions as is.

subdavis:
bump @zachmullen

PR #256, opened 2020-04-28
Bump 2.2.2

dchiquit:


package.json
```diff
@@ -128,4 +128,4 @@
       "stylelint-config-standard"
     ]
   }
-}
```
subdavis, line 131: don't remove newline :P

dchiquit: ordinarily my linter insists on newlines -_-

PR #259, opened 2020-06-01
Allow preUpload to update dest

jordansren:
When uploading files `dest` gets set before the `preUpload` function can make any changes to `dest`. This change will allow `preUpload` to potentially update `dest`, while keeping `dest` the same if `preUpload` does not make any updates to it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@girder/components",
-  "version": "2.2.2",
+  "version": "2.2.3",
   "scripts": {
     "serve": "vue-cli-service serve demo/main.js",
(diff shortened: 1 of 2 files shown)


zachmullen:
(Also this is an area where typescript would be super helpful in documenting the behavior of a user-provided function argument.)

subdavis:
@zachmullen Nope, this level of detail isn't documented.

Agree, a partial rewrite in typescript would provide a lot of value.

subdavis:
@jordansren please add a minor version bump in `package.json` to trigger an NPM release.

2.2.3 is appropriate, I think.

PR #277, opened 2020-10-06
Bind useful props to dropzone slot

zachmullen:
I've got a downstream app that is making use of a custom upload drop zone. Exposing these props to the slot makes things a lot more straightforward; without them I have to hack a few things.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@girder/components",
-  "version": "2.2.6",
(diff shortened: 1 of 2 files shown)


subdavis (commented):
I assume you'll want to deploy.  Might as well bump the package version in this PR, then I can approve?

zachmullen:
Yep, will do.

subdavis (approved):

PR #278, opened 2020-10-07
Properly sync browser state with route

zachmullen:


zachmullen:
This is a fair bit of complexity, but it provides a key optimization for those browsing the data hierarchy: paths and folder records don't have to be fetched on every navigation. Furthermore this is achieved without any sort of artificial caching. In a lot of cases, the information we need is already present in either the breadcrumbs or the list of child folders.

PR #283, opened 2020-11-18
Upload with limited worker pool

### LINKED ISSUE #282: Upper limit on parallel uploads
Caused by #231, now too many files cause the browser to freeze.

subdavis:
Limit upload concurrency to 5 requests.

This was pretty straightforward to change.

fixes #282

### DIFF
--- a/src/utils/mixins.js
+++ b/src/utils/mixins.js
@@ -249,9 +249,9 @@ const fileUploader = {
       // shallow copy into queue
       const fileQueue = this.files.slice().reverse();
 
-      /* Default to 5 workers */
-      const workerPool = [0, 1, 2, 3, 4];
-      await Promise.all(workerPool.map(async (workerId) => {
+      const WORKER_POOL_SIZE = 5;
+      const workerPool = [...new Array(WORKER_POOL_SIZE)];
+      await Promise.all(workerPool.map(async () => {
         let file = fileQueue.pop();
         while (file !== undefined) {
           // eslint-disable-next-line no-await-in-loop
@@ -260,7 +260,6 @@ const fileUploader = {
           }));
           file = fileQueue.pop();
         }
-        return workerId;
       }));
 
       const results = await Promise.all(promiseList);
--- a/src/utils/mixins.js
+++ b/src/utils/mixins.js
@@ -247,18 +247,17 @@ const fileUploader = {
       this.indeterminate = false;
       const promiseList = [];
       // shallow copy into queue
-      const fileQueue = this.files.slice().reverse();
-
+      const fileQueue = this.files.slice();
       const WORKER_POOL_SIZE = 5;
       const workerPool = [...new Array(WORKER_POOL_SIZE)];
       await Promise.all(workerPool.map(async () => {
-        let file = fileQueue.pop();
+        let file = fileQueue.shift();
         while (file !== undefined) {
           // eslint-disable-next-line no-await-in-loop
           promiseList.push(await this.uploadFile({
             file, hookResult, dest, uploadCls,
           }));
-          file = fileQueue.pop();
+          file = fileQueue.shift();
         }
       }));
 
--- a/src/utils/mixins.js
+++ b/src/utils/mixins.js
@@ -178,6 +178,59 @@ const fileUploader = {
       this.files = files;
     },
 
+    uploadFile({
+      file,
+      hookResult,
+      dest,
+      uploadCls,
+    }) {
+      let promiseChain = Promise.resolve();
+      if (file.status === 'done') {
+        // We are resuming, skip already completed files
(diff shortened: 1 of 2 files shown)


src/utils/mixins.js
```diff
@@ -178,6 +178,59 @@ const fileUploader = {
...
+            this.errorMessage = error.response.data.message || 'An error occurred during upload.';
+          } else {
+            this.errorMessage = 'Connection failed.';
+          }
+          file.status = 'error';
+          this.uploading = false;
+          this.$emit('error', { error, file });
+          throw error;
+        });
+    },
```
subdavis, line 232: This is a copy/paste from inside the promise map from L195.  No changes, just refactor.

src/utils/mixins.js
```diff
@@ -192,53 +245,24 @@ const fileUploader = {
...
-            parent: (hookResult && hookResult.dest) ? hookResult.dest : dest,
-            progress,
-            params: file.uploadClsParams,
-          });
-          promiseChain = promiseChain
-            .then(() => file.upload.beforeUpload())
-            .then(() => file.upload.start());
+      const promiseList = [];
+      // shallow copy into queue
+      const fileQueue = this.files.slice().reverse();
```
zachmullen, line 250: Any reason to bother reversing it?

subdavis: `.pop()` dequeues from the end, so to process them in order, you have to reverse the list first.

src/utils/mixins.js
```diff
@@ -192,53 +245,24 @@ const fileUploader = {
...
+
+      /* Default to 5 workers */
+      const workerPool = [0, 1, 2, 3, 4];
+      await Promise.all(workerPool.map(async (workerId) => {
```
zachmullen, line 254: Let's make this pool size a static constant.

```javascript
const WORKER_POOL_SIZE = 5;

const workerPool = [...Array(WORKER_POOL_SIZE)];
await Promise.all(workerPool.map(async () => {
```

Seems like we can get rid of `workerId` since it's not used.

zachmullen (changes requested):

src/utils/mixins.js
subdavis: Yeah, that's better.

src/utils/mixins.js
zachmullen: Ah, I had forgotten that the UI will be better if they go in order. Ideally we wouldn't even need to do a shallow copy or a reverse, and modifying the list each time isn't necessary -- we could just use a generator to iterate over the list and return each element in place. (If generators are too expensive to bother with, it could be an index variable in the closure...)

zachmullen: Or, if you think all of this bikeshedding isn't worth it, let's at least remove the `reverse()` and use `shift()` instead of `pop()`.

subdavis: JS has no "atomic test and set", so using an index variable requires you to have a little more confidence in how the JS thread scheduler works. 

`const file = queue.pop()` gives me more confidence than

```
while (i < this.files.length) {
  const file = this.files[i++];
}
```
because I'm not 100% sure that those lines won't be interleaved.  (I am like 90% sure).

> we could just use a generator 

A generator would just get transpiled with regenerator-runtime, right?  I'm pretty sure we're already bundling it, but after it gets transpiled, the memory footprint will probably be the same.

subdavis: Practically, I like `shift()`, but this discussion is also interesting, so if you want to make an argument for one of the other approaches I want to hear it.

subdavis:
Actually, I kinda need this to keep a customer happy.  Could we do `shift()` and then change it tomorrow if necessary?

src/utils/mixins.js
zachmullen: > because I'm not 100% sure that those lines won't be interleaved

Those lines will definitely not be interleaved. JavaScript is run via a single thread using an event loop approach for cooperative multitasking. The unit of atomicity is the function itself. Once a function starts, nothing will interrupt it until it reaches the end of its execution and returns. (The exception to this rule is if the function calls `await`.) So there's no need to ever worry about thread safety of data in JS. Even if you take advantage of web workers, which allows the use of multiple threads, there are very strict data ownership semantics that make atomicity concerns moot.

From a pure readability and simplicity standpoint, I find the generator approach best. It provides the exact behavior we want with no copying. But what I'm not sure on is how nasty that is from a babel / polyfilling perspective. Modern browsers support them natively now, but I'm assuming Babel will transpile them away with our configuration.

zachmullen:
> Actually, I kinda need this to keep a customer happy. Could we do shift() and then change it tomorrow if necessary?

Totally fine.

zachmullen (approved):

PR #284, opened 2020-11-19
Simplify upload worker pool code

zachmullen:
@subdavis this is a follow-up to #283 to use one of the alternative approaches I mentioned. I tried it using both the generator, and the index variable in closure, and ended up using the simple closure-based one. The generator one felt like overkill for the case of normal iteration over an array.

In addition to that change, I removed the unnecessary `promiseList` -- if I understand correctly, it was redundant since `await`ing the return value of `uploadFile` will already resolve to the result. Please correct me if there's something I'm missing there.

### DIFF
--- a/src/utils/mixins.js
+++ b/src/utils/mixins.js
@@ -245,23 +245,22 @@ const fileUploader = {
       this.errorMessage = null;
       const hookResult = await preUpload();
       this.indeterminate = false;
-      const promiseList = [];
-      // shallow copy into queue
-      const fileQueue = this.files.slice();
+      const results = [];
+      let i = 0;
       const WORKER_POOL_SIZE = 5;
       const workerPool = [...new Array(WORKER_POOL_SIZE)];
       await Promise.all(workerPool.map(async () => {
-        let file = fileQueue.shift();
-        while (file !== undefined) {
+        while (i < this.files.length) {
+          // must increment i before `await`, but after subscript operator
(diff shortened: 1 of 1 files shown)


src/utils/mixins.js
```diff
@@ -245,23 +245,21 @@ const fileUploader = {
...
+      const results = [];
+      let i = 0;
       const WORKER_POOL_SIZE = 5;
       const workerPool = [...new Array(WORKER_POOL_SIZE)];
       await Promise.all(workerPool.map(async () => {
-        let file = fileQueue.shift();
-        while (file !== undefined) {
+        while (i < this.files.length) {
+          const file = this.files[i];
+          i += 1;
```
subdavis, line 255:
```suggestion
          // eslint-disable-next-line no-plusplus
          const file = this.files[i++];
```
Unary is useful here.  I'd hate to have someone re-organize this function in such a way that breaks concurrency.  I don't know if I'd necessarily notice if someone accidentally moved Line 255 to the end of the loop, for example.

subdavis (changes requested):
This is nice.  One small comment.

src/utils/mixins.js
zachmullen: Sure, I will add a code comment too, otherwise it is definitely a threat of someone thinking they can make this a for loop, or move the increment around.

subdavis (approved):
:+1:

PR #288, opened 2020-11-30
Add some flair to the upload drop zone

zachmullen:
I've been staring at the upload component a lot lately in a few different app contexts, and I just felt like it needed a little bit more of a call to action when it was in the waiting for files state. I've seen some other drag and drop upload UIs that show a little frame around the corners of the drop zone, and tried it out here. Let me know what you think.

Screen Shot 2020-11-30 at 4 32 56 PM
Screen Shot 2020-11-30 at 4 33 12 PM

### DIFF
--- a/src/components/Presentation/Dropzone.vue
+++ b/src/components/Presentation/Dropzone.vue
@@ -6,6 +6,7 @@
     @dragleave="dropzoneClass = null"
     @drop="dropzoneClass = null"
   >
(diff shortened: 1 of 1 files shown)


subdavis (approved):
Love it.

zachmullen:
It's up on https://gwc.girder.org/ if anyone wants to test manually.

PR #289, opened 2020-12-01
Add transition between upload dropzone and file list

zachmullen:
This addresses one other qualm I've been having with the upload component, which was that it was visually jarring when switching between the dropzone and the files list. A simple transition makes this a lot better IMO (GIF framerate probably doesn't do it justice).

uploader


subdavis (approved):
:+1:

PR #297, opened 2021-01-05
Better document vue/vuetify installation

subdavis:
Made both Vue and Vuetify peer dependencies.

Had to change the install guide to explicitly install vuetify, and realized I was just replicating Vuetify docs inline.

So I switched to recommending `vue add vuetify` as the install mechanism, since it does quite a lot of work that a user would otherwise have to guess from looking at the GWC package json and vue.config.js

This way, for the most common case of "fresh project", you get a more first-class vuetify experience rather than trying to replicate our config by hand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,25 +5,33 @@ It works with [data.kitware.com](https://data.kitware.com/).
 
 ## Usage Quickstart
 
+**This library only supports Vue 2.**
+
 ### Installation
 
+[Vue CLI](https://cli.vuejs.org/) is required for building applications with Girder web components.  You will also need to install [Vue CLI Plugin Vuetify](https://next.vuetifyjs.com/en/getting-started/installation/#vue-cli-install).
+
+First, install Vuetify if you aren't already using it.
+
+``` bash
+vue add vuetify
+```
+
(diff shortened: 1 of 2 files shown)


jbeezley (dismissed):

zachmullen (dismissed):

subdavis:
The issue is installing vue/vuetify for running tests and building demo app.

Doing this is [unsupported by yarn](https://github.com/yarnpkg/yarn/issues/1503).  After 4 years of debate, a maintainer has decided "you can do it with a plugin in yarn 2".   There are [weird packages](https://dev.to/yvonnickfrin/how-to-handle-peer-dependencies-when-developing-modules-18fa) to hack around it, but I don't feel good about using one.

The proper yarn 1.X solution would be either a) a monorepo or b) an entirely separate repo for the demo app, neither of which is appropriate for the level of effort involved.

I think my docs changes are still good, but otherwise, not doing this.

PTAL.

subdavis:
@zachmullen this is the last thing blocking the 3.0 chain of events.  Mostly just because I want the NPM page to include these changes.

zachmullen (approved):

Issue #299, opened 2021-01-07
SSRF vulnerability in axios version < 0.21.1


scottwittenburg:
Our project has a transitive dependency on `axios` via this project.  A couple days ago, we started seeing high severity dependabot alert regarding axios versions `< 0.21.1`.

I can make a PR simply updating the version, but as a lot of folks have started moving to the latest version of `axios`, at least one report came in about some features maybe not working the same after moving from `0.19` to `0.21`.  I have no idea whether this project would be affected by that, or if transition would be smooth.

Here's the original `axios` [issue](https://github.com/axios/axios/issues/3407), and a link to the more recent [issue](https://github.com/axios/axios/issues/3526) created after people started picking up the fix.  

Please let me know what I can do to help.

zachmullen:
We are on axios v0.21.1 in this library.

zachmullen:
@scottwittenburg if you want to create a new release in this repo, feel free and I'll publish it.

scottwittenburg:
Ah, sorry for the noise, I thought I had checked the version you were using.

What would I do to create a new release?

Thanks.

zachmullen:
Just go into `package.json` and bump the version from 2.3.0 to 2.3.1, and make a PR of that. I'll approve that and merge it, then push a git tag, which will trigger the publication of the new package. Thanks!

scottwittenburg:
Done in #300, thanks @zachmullen!

PR #300, opened 2021-01-07
v2.3.1

scottwittenburg:


zachmullen (approved):

zachmullen:
Tag has been pushed. 2.3.1 should be published in a few minutes.

scottwittenburg:
> Tag has been pushed. 2.3.1 should be published in a few minutes.

Thanks for the lightning fast action @zachmullen 😁

subdavis:
@scottwittenburg @zachmullen 2.3 was an unintended release, and I was planning to scrub it from NPM pending a discussion.

The changes introduced in #293 are significant enough to require a major version update, so I was going to suggest moving to 3.0.0 after https://github.com/girder/girder_web_components/pull/297 gets resolved, which I haven't had time to fix.

Do you need the changes introduced in 2.3.0 (ts, import restructure)?

We can certainly keep going with 2.3, but it would somewhat contradict semver guidelines.

subdavis:
I see, the point was to release the changes from #299.

@scottwittenburg you should know that upgrading your project to 2.3 will break your project.

This library hasn't had so much traffic that it has been necessary to maintain a "stable" branch and a "next" branch to allow us to push minor patches like #299 while working on larger changes.

Apologies for the confusion.  Let me know what you want do now.

zachmullen:
I don't have a strong preference.

scottwittenburg:
@subdavis We're currently on `2.2.6` and we're only using only the user login/registration dialog in our application, I don't think we necessarily need anything from `2.3`, other than the newer `axios`.  And based on your last comment, it sounds like getting that change will require some other work.

Other than that, I don't have any preference which path you take.  But when you decide what you're going to do, I'll get in touch about the path forward for my application.  Thanks!

subdavis:
Unless someone has a better suggestion, I'm going to

* Get #297 merged
* Release 3.0.0
* Scrub 2.3.0, 2.3.1
* Write a short migration guide (should be easy, just have to change import paths)

scottwittenburg:
> * Write a short migration guide (should be easy, just have to change import paths)

That would be great, thanks!

PR #304, opened 2021-03-15
Typescript/fix defs

subdavis:
Typescript fixes for my needs on other projects.  Apparently imports must happen outside the declare block.

Added `[key: string]: any;` to give downstream more flexibility to use the models.


zachmullen (approved):

subdavis:
I misunderstood a lot about how module definitions work.  In order to declare different modules, they have to be in different files.

Also, that `export {}` chunk isn't correct either.  I thought I tested this with a real project before 3.0.0 but apparently I was just guessing.

zachmullen (approved):

PR #307, opened 2021-03-29
Show better error for email verification required

### LINKED ISSUE #306: Show better error for email verification required
> UI for Girder itself shows the user an error message saying email verification is required and provides a link to resend the verification request email.  The GirderAuthentication component just shows the error message indicating email verification is required, but does not provide a link to resend the verification request email.  I don't see any props on the component that look promising for customizing this behavior, so is the only option for the user to try logging in through Girder itself?

subdavis:
Screenshot from 2021-03-29 16-08-26

Screenshot from 2021-03-29 16-09-22

fixes #306

### DIFF
--- a/src/components/Authentication/Authentication.vue
+++ b/src/components/Authentication/Authentication.vue
@@ -35,11 +35,12 @@
 </template>
 
 <script>
+import Vue from 'vue';
 import GirderLogin from './Login.vue';
 import GirderRegistration from './Register.vue';
 import { OauthTokenPrefix, OauthTokenSuffix } from '../../rest';
 
-export default {
+export default Vue.extend({
   inject: ['girderRest'],
   components: {
     GirderLogin,
@@ -101,5 +102,5 @@ export default {
       }
     },
   },
-};
+});
 </script>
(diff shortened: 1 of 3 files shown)


src/components/Authentication/Authentication.vue
```diff
@@ -35,11 +35,12 @@
 </template>
 
 <script>
+import Vue from 'vue';
 import GirderLogin from './Login.vue';
 import GirderRegistration from './Register.vue';
 import { OauthTokenPrefix, OauthTokenSuffix } from '../../rest';
 
-export default {
+export default Vue.extend({
```
zachmullen, line 43: What's this change for?

subdavis: I make that change in any SFC I touch that doesn't have it when I remember.   It enables VS Code's Vetur plugin to provide template interpolation, so it can do basic type checking and warn about missing variables / typos/ / etc in the template.

It has no impact on the bundle, `vue-loader` feeds the default export to the same `Vue.extend()` function during build.

zachmullen (approved):

PR #310, opened 2021-04-08
Search theme fix

subdavis:
See https://github.com/vuetifyjs/vuetify/issues/2792 for why I did the `Math.random()` hack.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@girder/components",
-  "version": "3.0.2",
+  "version": "3.0.3",
   "scripts": {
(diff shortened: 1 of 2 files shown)


zachmullen (approved):

BryonLewis:
Quick search, while your at it? https://github.com/girder/girder_web_components/blob/3208eb8ca7c72b4df7089ea45a75d70d2ade695d/src/components/AccessControl.vue#L200  - minor access control v-select defaulting to light theme as well.

BryonLewis:
ahhh too late.

PR #326, opened 2021-09-28
Items per row sync

BryonLewis:
Implementing itermPerRow sync ability to send information about the current `intialItemPerRow` value as the user changes it in the v-data-table dropdown.  This is needed for a DIVE feature request where this value is saved on the client and remembered after swapping from different pages.  Just additional signals so shouldn't interfere with existing usage of GWC.


subdavis (changes requested):
If the variable updates, it is no longer an initialization value, so the name doesn't make sense.

Perhaps this should just be `itemsPerPage` now?  I think the prop just needs to change.  It's a relatively minor breaking change that we can mention in the release log.

Please also put a minor version upgrade into the package json

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@girder/components",
-  "version": "3.0.4",
```
subdavis, line 3: Minor version upgrade should reset patch to 0 i.e. `3.1.0`

src/components/DataBrowser.vue
```diff
@@ -93,6 +93,9 @@ export default {
         this.$emit('input', val);
       },
     },
+    itemsPerPageInternal() {
+      return this.options.itemsPerPage;
+    },
```
subdavis, line 98: What is the point of this computed property?

`'options.itemsPerPage'` works as a watcher key.

https://vuejs.org/v2/api/#watch

subdavis (changes requested):

src/components/DataBrowser.vue
BryonLewis: I knew that at one point and for some reason just forgot about it.  It's interesting what your brain does to work around stuff when you forget about specific pieces of information.

src/components/snippets/FileManager.vue
```diff
@@ -200,8 +200,9 @@ export default Vue.extend({
       :draggable="dragEnabled"
       :root-location-disabled="rootLocationDisabled"
       :value="value"
-      :initial-items-per-page="initialItemsPerPage"
+      :items-per-page="itemsPerPage"
       :items-per-page-options="itemsPerPageOptions"
+      @update:itemsPerPage="$emit('update:itemsPerPage', $event)"
```
subdavis, line 205: does `:items-per-page.sync` work when you use pascalCase in the emitter?

Screenshot from 2021-09-28 15-46-34

If so, I think this is fine to leave as-is.

subdavis (approved):

src/components/snippets/FileManager.vue
BryonLewis: yeah in my testing I was using ke-bab case `:items-per-page.sync` inside of DIVE.  Should I unify them?  I just feel weird calling `$emit('update:this-test-variable')` and having `thisTestVariable` be non-searchable in the component file because it is in ke-bab case for the emit instead of camelCase.

subdavis: nah, it's cool like it is.

PR #336, opened 2022-03-15
Row slot

subdavis:
## Changes

The old slot was too narrowly scoped.  Rather than maintain backward compatibility and have 2 named slots in exactly the same place, this introduces a breaking change:

* Remove `row-widget` slot in DataBrowser, DataTable, FileManager
* Replace with `row` slot that can be used to customize the entire presentation of the item/folder row.

## Release

Release 3.2.0

## Example Usage

Trivial example to make the text italicized.

``` html
<girder-file-manager>
    <template #row="props">
      <i>{{ props.item.name }}</i>
     </template>
</girder-file-manager>
```

### DIFF
--- a/demo/App.vue
+++ b/demo/App.vue
@@ -214,7 +214,15 @@
                 :upload-multiple="uploadMultiple"
                 :upload-enabled="uploadEnabled"
                 @dragend="dragend"
-              />
+              >
+                <!--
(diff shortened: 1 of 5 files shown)


zachmullen (approved):

cjh1:
Awesome!

PR #360, opened 2026-02-18
ci: fix gwc.girder.org

finetjul:
Contrary to Webpack, Vite does not process a .js file as input but HTML files. By default, Vite considers index.html located at the root of the project.


Justineantoine:
LGTM

Justineantoine (approved):

